PR #19, opened 2019-06-20
Bitset refactor

jeremiedb:
Move to BitSet histogram approach. 
Significant speedups for limited number of bins (< 64). NBins grater than 100 may result in slowdown. 
To do: add option to switch between the loop over bins or add heuristic to emulate high number of split points (by building N sets of bags for the equivalent of N*nbins behavior).


jeremiedb:
@JuliaRegistrator register()

PR #42, opened 2020-02-17
Add Manifest.toml in gitignore

tlienart:
I should have done that in #41, sorry


jeremiedb:
I've just pushed another commit making that change #43  
Thanks for the fixes, it now builds successfully on Julia 1.0 - 1.3.

PR #60, opened 2020-11-20
Extend [compat] for Distributions and add CompatHelper

ablaom:
- Adding Distributions 0.24 compatibility
- Adding compat helper which automatically generates PR's when dependencies release new versions that are outside the current bounds.


ablaom:
Not sure how CI works here but tests pass locally.

@jeremiedb

jeremiedb:
Thanks for this, I wasn't aware of the CompatHelper, looks very convenient!
Seems like I had neglected the travis.org migration, I think it should now be back.

ablaom:
Don't forget to tag a new release 😄

PR #70, opened 2021-02-02
Update MLJModelInterface & add gini eval metric

jeremiedb:


src/MLJ.jl
```diff
@@ -21,6 +21,9 @@ function okay_to_continue(new, old)
     new.metric ==  old.metric
 end
 
+MLJModelInterface.reformat(::EvoTypes, X, y) = (MLJModelInterface.matrix(X), y)
```
ablaom, line 24: If you implement this, then you need to change your `fit` method correspondingly. It now receives the model-specific representation of data (the output of your `reformat` method).

Also, you should define a `reformat` and `selectrows` with a signature matching the `predict` method and then, similarly, change your `predict` to expect the model-specific representation of data. 

Docs are here: https://github.com/alan-turing-institute/MLJ.jl/blob/mlj2/docs/src/adding_models_for_general_use.md#implementing-a-data-front-end

Hope this resolves your build fail. Happy to provided further guidance here. 

Note that MLJ itself is not synched with the data front-end yet, although latest MLJBase is.

ablaom: And I suggest you tag your new release as a minor version (breaking); see the docs referenced above.

jeremiedb: @ablaom Thanks! I think the aspect of `fit` and `predict` input type is fine, as everything still assumes the `Matrix` format. 
Actually, the only pain point seems to be with CategoricalArrays (another reason I always work straight on Floats or Ints :) ). 
The Issue stems from the fact that a view of a `CategoricalVector` isn't a subtype of `AbstractCategoricalVector` :\
I got it fixed to use a condition on `eltype` to be a `CategoricalValue`. 

Also with the classification, there was a warning on the predictions if the pool was not specified. It was settled with `pool=missing`. Should it be a default setting for pool in `UnivariateFinite`?

src/MLJ.jl
```diff
@@ -46,7 +46,7 @@ end
 function predict(::EvoTreeClassifier, fitresult, Xnew)
     Xnew = MLJModelInterface.matrix(Xnew)
     pred = predict(fitresult, Xnew)
-    return MLJModelInterface.UnivariateFinite(fitresult.levels, pred)
+    return MLJModelInterface.UnivariateFinite(fitresult.levels, pred, pool=missing)
```
ablaom, line 49: This does not seem right. I would be happy to look into this but can you please just merge #69 and make this a separate exercise. The compat issue is holding things up at our end. We want to release a new MLJModels/MLJ but it's tricky if your package is not up-to-date (because of way model registry is updated.

Thanks.

ablaom: To be clear: there is no need for you to buy into new data front-end to use MLJModelInterface 0.4

ablaom: Also in the meantime, you may want to look at how the pool business is taken care of for XGBoost.jl:

https://github.com/alan-turing-institute/MLJXGBoostInterface.jl/blob/325c22d3875d9b84540138817660aeb9b7f842fc/src/MLJXGBoostInterface.jl#L589

https://github.com/alan-turing-institute/MLJXGBoostInterface.jl/blob/325c22d3875d9b84540138817660aeb9b7f842fc/src/MLJXGBoostInterface.jl#L622

https://github.com/alan-turing-institute/MLJXGBoostInterface.jl/blob/325c22d3875d9b84540138817660aeb9b7f842fc/src/MLJXGBoostInterface.jl#L643

https://github.com/alan-turing-institute/MLJXGBoostInterface.jl/blob/325c22d3875d9b84540138817660aeb9b7f842fc/src/MLJXGBoostInterface.jl#L660

src/MLJ.jl
ablaom: > Also with classification, there was a warning on the predictions if the pool was not specified. It was settled with pool=missing. Should it be a default setting for pool in UnivariateFinite?

What is the warning? The pool should not be missing. The first argument `levels` to `UnivariateFinite` constructor is ordinarily a CategoricalArray and the pool is extracted from that.

Also, it's strange this warning is only appearing now. What is new here?

ablaom: > Thanks! I think the aspect of fit and predict input type is fine, as everything still assumes the Matrix format.

I don't understand this statement. The output of the `reformat` that is defined in this PR is (X, y) where X is a matrix. But your`fit` method starts by transforming it's `X` argument into a matrix, as if it were a table. But it won't be a table anymore, now that you have overloaded `reformat`. It will already be a matrix. That's kind of the point.

jeremiedb: Regarding the pool warning during prediction:

```
┌ Warning: No `CategoricalValue` found from which to extract a complete pool of classes. Creating a new pool (ordered=false). You can:
│  (i) specify `pool=missing` to suppress this warning; or
│  (ii) use an existing pool by specifying `pool=c` where `c` is a `CategoricalArray`, `CategoricalValue` or CategoricalPool`.
│ In case (i) specify `ordered=true` if samples are to be `OrderedFactor`.
```

So I followed the warnings in specifying the pool as missing. It's a good point though regarding what's new, I couldn't exclude that the warning was already present and just has been simply ignored. 

Anyhow, all tests are now passing and no warnings, so I would move forward merging and releasing this PR. 
And sorry about the delay, I hadn't realize I was bottlenecking the MLJ release process.

jeremiedb: > I don't understand this statement. The output of the `reformat` that is defined in this PR is (X, y) where X is a matrix.

My intent was only to clarify that the input type wasn't the cause of the failing tests. You're right that the additional matrix conversion of `X` would now be redundant. I'll remove it, though for now it doesn't hurt performance and I don't want to postpone more the PR!

ablaom: I looked into the levels business. Your choice to add `pool=missing` is legitimate after all and I was wrong. (Another solution is to change the line https://github.com/Evovest/EvoTrees.jl/blob/280778c096add04fb2d0fba4ee46d82ed18c0b08/src/fit.jl#L16 to

```
levels = MLJModelInterface.classes(first(Y))
```

but it amounts to the same thing and is probably less readable from your point of view.)

I double checked that your implementation tracks unseen classes appropriately (they even get a non-zero probability - I guess you have some kind of regularisation going on here). 

So all good with the classes.

ablaom:
> I'll remove it, though for now it doesn't hurt performance and I don't want to postpone more the PR!

Your call, but I'd remove that one line now. It's a bit confusing. After all, I will be using your code as a exemplar for the reformat buy-in 😉 .

Thanks for your cooperation on this!

ablaom:
Just FYI: To implement data front-end on the predict side, add:

```julia
MLJModelInterface.reformat(::EvoTypes, X) = (MLJModelInterface.matrix(X), )
MLJModelInterface.selectrows(::EvoTypes, I, A) = (view(A, I, :), )
```
and delete all the `MLJModelInterface.matrix(X)` lines in your predicts.

jeremiedb:
done with the redundant Matrix conversion for both `fit` and `predict`. EvoTrees v0.6.0 is now ready on General Registry.

PR #75, opened 2021-03-12
Add feat importance into MLJ report

jeremiedb:
Fix for https://github.com/alan-turing-institute/MLJ.jl/issues/745
Since model input is a matrix, it doesn't carry the variable names, so the report is formed by pairs in the form
["feat_1" => 0.24,
"feat_5" => 0.05, ...]


ablaom:
> Fix for [alan-turing-institute/MLJ.jl#745](https://github.com/alan-turing-institute/MLJ.jl/issues/745)
> Since model input is a matrix, it doesn't carry the variable names, so the report is formed by pairs in the form
> ["feat_1" => 0.24,
> "feat_5" => 0.05, ...]

This seems a pity, as the names are indeed available to the mlj wrapper, since the data comes in as a table, no?

PR #98, opened 2021-04-22
Dev

jeremiedb:


jeremiedb:
@ablaom I'll make a fix release with the patch you provided. 
Would you have any objection to move MLJModelInterface requirement to 1.0+ rather than the current 0.3+?

ablaom:
According to the release notes there has been nothing breaking in MLJModelInterface for model providing packages since   0.3; and 1.0 is identical to 0.4.1. I'd suggest you leave the old versions there unless you have a reason not to (I've done this in all the other interfaces). But it's your call 😄

Issue #91, opened 2021-04-14
Merge with MemoryConstrainedTreeBoosting.jl?


brianhempel:
Maybe we can merge projects?

EvoTrees:
- More loss functions supported
- Cleaner code
- Working on integrating with Julia ecosystem

[MemoryConstrainedTreeBoosting](https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl):
- 4-5x faster on CPU
- Early stopping

MemoryConstrainedTreeBoosting.jl is a library I've been working on for a couple years that would allow me to control the loading and binning of data, so I could (a) do feature engineering in Julia and (b) use all the memory on my machine for the binned data. I've also spent a lot of time on speed because 10% faster ≈ training done 1 day sooner for my data sets. I think it is quite fast. I didn't bother documenting the library until today, however.

With our powers combined...!

A benchmark below. 4 threads on my 2013 quad-core [i7-4960HQ](https://www.cpu-world.com/CPUs/Core_i7/Intel-Core%20i7-4960HQ%20Mobile%20processor.html).

---

```
 pkg> add https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl
```

```julia
using Statistics
using StatsBase:sample
using Revise
using EvoTrees

nrounds = 200

# EvoTrees params
params_evo = EvoTreeRegressor(T=Float32,
        loss=:logistic, metric=:logloss,
        nrounds=nrounds,
        λ=0.5, γ=0.0, η=0.05,
        max_depth=6, min_weight=1.0,
        rowsample=1.0, colsample=0.5, nbins=64)

# MemoryConstrainedTreeBoosting params
params_mctb = (
        weights                 = nothing,
        bin_count               = 64,
        iteration_count         = nrounds,
        min_data_weight_in_leaf = 1.0,
        l2_regularization       = 0.5,
        max_leaves              = 32,
        max_depth               = 6,
        max_delta_score         = 1.0e10, # Before shrinkage.
        learning_rate           = 0.05,
        feature_fraction        = 0.5, # Per tree.
        bagging_temperature     = 0.0,
      )

nobs = Int(1e6)
num_feat = Int(100)
@info "testing with: $nobs observations | $num_feat features."
X = rand(Float32, nobs, num_feat)
Y = Float32.(rand(Bool, size(X, 1)))


@info "evotrees train CPU:"
params_evo.device = "cpu"
@time m_evo = fit_evotree(params_evo, X, Y);
[...]

PR #100, opened 2021-04-30
25% speedup PoC w/SIMD, double histogram building, and observation chunking

brianhempel:
Continuing #91.

Only on Float32 gradient loss for now. The `SIMD.Vec{Float64,4}` type should work (and trigger AVX instructions) but somehow it crashed for me, so there's duplicate non-SIMD code for Float64 :(. A bit of fighting should work that out but this is PoC. 
Worse case scenario we have to use `SIMD.Vec{Float64,2}`.

The code for the other types of loss should be similar. The δ𝑤's already have the appropriate padding so you can load them in blocks of 4/8 floats.

See the commit messages for the various changes.

There's probably some extra allocations because of using `Iterators.partition` to chunk the indices instead of [doing the math by hand](https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl/blob/398938bd52a74bc026d1731363ff4207514b2933/src/MemoryConstrainedTreeBoosting.jl#L1687) but that's another easy win for later.

Can also try chunking features in groups of 8 or 12.

I'm sure the PoC breaks the GPU code, sorry. I don't think I have libcuda.


jeremiedb:
How did you make the comparison for the performance? 
Running from my laptop (Ryzen 4800 16 threads), `dev-refactor` branch shows a performance about 10-%20% faster and less allocations (in RAM, not numbers though) than for this PR on the [benchmark v3](https://github.com/Evovest/EvoTrees.jl/blob/master/experiments/benchmarks_v3.jl). 

This PR: 
```julia
julia> @time fit_evotree(params_evo, X, Y);
 18.801873 seconds (197.64 k allocations: 9.835 GiB, 4.54% gc time)
```

[`dev-factor`](https://github.com/Evovest/EvoTrees.jl/tree/dev-refactor)
```julia
julia> @time fit_evotree(params_evo, X, Y);
 16.011239 seconds (1.07 M allocations: 2.579 GiB, 1.74% gc time)
```

Do you observe such regression on your side?

brianhempel:
I rebased on your work and still obtain similar results to what I claimed.

My tests are versus https://github.com/brianhempel/EvoTrees.jl/commit/4d41a81c9813ced182e473ed9b6a2d4be1ff0a3f and are run on a quad-core Haswell processor w/Julia 1.5.4. I see similar speedups vs https://github.com/brianhempel/EvoTrees.jl/commit/4d41a81c9813ced182e473ed9b6a2d4be1ff0a3f on Julia 1.6.1 but Julia 1.5.4 is still ~10% faster. 9gb allocation? Wild. I only see 3gb, over 600k allocations, copy-pasting [this](https://github.com/Evovest/EvoTrees.jl/blob/master/experiments/benchmarks_v3.jl#L4-L41) straight into Julia:

This PR, brianhempel/dev-refactor:
```
$ JULIA_NUM_THREADS=4 julia-1.6 --project=.
...
julia> @time fit_evotree(params_evo, X, Y);
 23.329188 seconds (595.37 k allocations: 2.941 GiB, 1.71% gc time)
```

origin/dev-refactor:
```
$ JULIA_NUM_THREADS=4 julia-1.6 --project=.
...
julia> @time fit_evotree(params_evo, X, Y);
 31.123589 seconds (578.78 k allocations: 2.523 GiB, 0.90% gc time)
```

Yes, there is a small increase in allocation amount because there's dummy space now so the SIMD can work.

Can you confirm the benchmarks you are running are 1e6 obs, 100 features? It is also possible that one of the changes is friendly to Intel-only. I guess I'd be a little surprised though, Haswell is pretty old. I'm sure AMD would have caught up on features to Haswell by now. But MCTB is faster on your machine, right? MCTB is just a similar series of optimizations, albeit also with a [consolidation of dw](https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl/blob/398938bd52a74bc026d1731363ff4207514b2933/src/MemoryConstrainedTreeBoosting.jl#L1276) before building the histogram for each leaf, to reduce cache fetches/misses. Also MCTB chunks the features as well and uses work stealing, although work stealing didn't help this benchmark so I didn't include it.

"Padding in prep for SIMD" should be a regression but each commit after that should be an improvement. I don't know if you have the patience to go commit-by-commit but that might help. I'm still miffed on the 9gb allocation.

jeremiedb:
I might have ran the `master` rather than the `dev-refactor` branch for the benchmarks. I reran on another machine (AMD Threadripper using 16 threads) and now got a ~15% improvement on this PR compared to current dev-refactor branch, and memory usage are now more aligned. 

Using the latest [`dev-refactor branch`](https://github.com/Evovest/EvoTrees.jl/tree/dev-refactor) : 
```julia
[ Info: evotrees train CPU:
  2.655647 seconds (4.00 M allocations: 1.513 GiB, 4.07% gc time, 48.82% compilation time)
 29.636819 seconds (81.12 M allocations: 7.055 GiB, 2.69% gc time)
 12.952900 seconds (1.01 M allocations: 2.559 GiB, 1.19% gc time)

# MCDT
 2.849717 seconds (3.72 M allocations: 555.577 MiB, 2.09% gc time)
```

Using your [`dev-refactor branch`](https://github.com/brianhempel/EvoTrees.jl/tree/dev-refactor): 

```julia
[ Info: evotrees train CPU:
  2.835524 seconds (4.01 M allocations: 1.522 GiB, 8.84% gc time, 34.63% compilation time)
 27.946466 seconds (81.88 M allocations: 7.517 GiB, 2.84% gc time)
 11.776919 seconds (1.07 M allocations: 2.985 GiB, 2.05% gc time)

# MCTB
 2.999016 seconds (3.73 M allocations: 556.023 MiB, 1.91% gc time)
```

What puzzles me is even if I let away 2 secs for EvoTrees initialisation, there would still be a huge 3X gap with MCTB. Did you observe such magnitude in speedup moving from "vanilla" histograms to the SIMD version? It seems like there's something more fundamental to the performance that is missing in this PR vs MCTB. Maybe isolating the histogram benchmark in a test such as [this one](https://github.com/Evovest/EvoTrees.jl/blob/29126fcf59a5aae4439ca06f39232b1fd7061e5a/experiments/hist_v1.jl#L174-L195) might help to confirm. 

Otherwise, there's the resampling of the obersations that is triggered at each iterations in EvoTrees that eats up some resources and whch is not present in MCTB. Yet, the observed fitting process timing of 
EvoTrees looks roughly aligned with what the timing observed for the histogram update. 
No other feature of MCTB such as `bagging_temperature` that may affect performance or shrink the search space during construction?

brianhempel:
Just to confirm: depth 6 is 16 leaves, 15 internal nodes (leaves count in the depth)?

Rather than making separate benchmarks, I just throw `@time` in a suspected spot in my code to eyeball. For example, 0.3s out of EvoTree's 23s on my machine is the obs subsampling, which can be elided if the subsample fraction is 1. Bagging is the similar operation in MCTB, but bagging uses all the obs all the time, it just re-weights them as if they were being sampled.

- 0.8s is `predict!`.
- 2s is `update_grads!`. This could be threaded. It also doesn't need to call sigmoid: sigmoid and logloss are complementary, so the grads work out to be [extremely simple](https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl/blob/7aeaed385cd68335852bcc9d949c26a8a7f07b34/src/MemoryConstrainedTreeBoosting.jl#L971-L975). (I checked vs XGBoost and worked it out by hand too.)
- Only half the time (10-15s) is `update_hists!`, which is surprising. I would expect it to be more.
- 0.2s is `update_gains!`
- 1s is `split_set_threads!`. I would try `Threads.@threads` instead of `Threads.@spawn`. I think the latter is more book-keeping (it spawns a Task), and its use might explain why more threads increases the number of allocations significantly.

I see that you only use histogram subtraction for every other node. For MCTB, I calculate the smallest node first to do the least amount of work, then the sibling is computed via subtraction. That makes a big difference when the dataset is imbalanced (not likely the case for this benchmark, but true for my dataset). So I'd recommend that, too. Also, the histogram subtraction can be threaded.

MCTB has been optimized with a fine-toothed comb. I'm happy to list all the tricks when it comes to it. But maybe try the above and poke around with `@time` to find more low-hanging fruit.

jeremiedb:
Regarding the basis of the predictor, I did have a preference for the smoothness of using the linear predictor of the logistic function. However, computationnally, it appears to add quite a burden. I compared the running time with using the `linear` predictor and it dropped by nearly 5 secs! So I'll switch to that probability basis. 
 
> Just to confirm: depth 6 is 16 leaves, 15 internal nodes (leaves count in the depth)?

Depth 6 contains 63 total nodes in EvoTrees (31 splits and 32 preds). I guess this above is a typo, as I looked at one of the trees in MCTB and I also counted 31 splits nodes and 32 leaves.  

I didn't get improvements using a naive multi-threaded `update_grads!` with `@inbounds @threads for i in eachindex(y)`, but here I guess it should be launched in num_threads partitions given the low cost of each gradient calculation. 

The `eval` metric calculation also adds another 1 to 1.5 sec as showned when `metric=:none` instead of `metric=:logloss`. Not sure if using a trick similar to what is used on the gpu side would help, that is perform a multi-threaded calculation in a pre-allocated vector of evals followed with a straight `mean`. 

I got no allocations improvement from `@threads` rather than `@spawn`, but appears like having a slight overhead advantage, so I'll switch to `@threads` (and coherent with it also being used for histograms update). 

I like the idea of identifying the smallest child node for histogram update. I'll give it a shot. 

Thanks again for your feedback. With all the above, I think it should help cover a decent ground of the low haging fruits.

brianhempel:
> typo

👍  I miscounted when I inspected a tree, sorry!

> Regarding the basis of the predictor, I did have a preference for the smoothness of using the linear predictor of the logistic function. However, computationally, it appears to add quite a burden. I compared the running time with using the linear predictor and it dropped by nearly 5 secs! So I'll switch to that probability basis.

I misspoke: you still need to do [one sigmoid to get ŷ](https://github.com/brianhempel/MemoryConstrainedTreeBoosting.jl/blob/7aeaed385cd68335852bcc9d949c26a8a7f07b34/src/MemoryConstrainedTreeBoosting.jl#L1048-L1053), so you should get benefits from threading if there are 1e6 obs. I guess you were relying on the optimizer to combine your sigmoids. You might have to check to make sure that actually happened.

Maybe I'm not up on my stats, but I'm not sure what you mean by smoothness. If the pre-sigmoid "margin" (called "score" in MCTB, the logit value), then:

d/dm logloss(y, σ(m))   =   σ(m) - y

d²/dm² logloss(y, σ(m))   =   σ(m) * (1 - σ(m))

That's smooth to me! Also, that second derivative is the same as in EvoTrees. Not sure what was going on with the first derivative, maybe it simplifies out to the same.

jeremiedb:
Good, I think I misinterpreted your comment as having the raw prediction in the probability domain (0, 1) rather than in the logit domain (-inf, inf). This is what I was referring through the "non-smothness" as the logloss isn't continuously over Reals if x represents the probability rather than the logit transformation. As for the first derivative, it was indeed equivalent, I've fixed it to the simplified form. 
The pure training process as defined by `EvoTrees.grow_evotree!` now appears roughly 2X slower than MCTB. I'll have limited time in the upcoming days to push that further, but I now have some clearer visibility on where to improve.

PR #140, opened 2022-03-22
GPU: Annotate scalar indexing with `CUDA.@allowscalar`

DilumAluthge:


DilumAluthge:
To identify these locations, I set `CUDA.allowscalar(false)` at the top of my script, and then I trained some example models on the GPU.

jeremiedb:
Thanks for this PR! 
It will also be useful to highlight potential bottlenecks in the GPU implementation as its performance still lags behind that of xgboost.

Issue #166, opened 2022-09-29
Update MLJ feature importance access to current standard


john-waczak:
Currently feature importances from EvoTrees are accessed via `report()`. We should update the interface to comply with the new method defined here: https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Feature-importances. In particular I suggest we do the following: 
- remove automatically computed feature importances from `report()`
- dispatch `MLJModelInterface.feature_importances(model::M, fitresult, report)` to compute feature importances *as desired*
- dispatch `MLJModelInterface.reports_feature_importances(::Type{<:M}) = true` for each EvoTrees model so MLJ knows we can access feature importances.

PR #167, opened 2022-09-29
Updated MLJ interface for feature importances per #166

john-waczak:


ablaom:
@OkonSamuel

src/MLJ.jl
```diff
@@ -45,7 +46,8 @@ function MMI.update(model::EvoTypes, verbosity::Integer, fitresult, cache, A, y)
     grow_evotree!(fitresult, cache)
   end
 
-  report = (feature_importances=importance(fitresult, A.names),)
+  # report = (feature_importances=importance(fitresult, A.names),)
+  report = ()
```
jeremiedb, line 50: Shouldn't it return the feature names, same as in the above `fit` function? 
`report = (features=A.names,)`

src/MLJ.jl
```diff
@@ -5,7 +5,8 @@ function MMI.fit(model::EvoTypes, verbosity::Int, A, y)
     fitresult, cache = init_evotree(model, A.matrix, y)
   end
   grow_evotree!(fitresult, cache)
-  report = (feature_importances=importance(fitresult, A.names),)
+#  report = (feature_importances=importance(fitresult, A.names),)
```
jeremiedb, line 8: I'd remove the commented out line.

src/MLJ.jl
```diff
@@ -45,7 +46,8 @@ function MMI.update(model::EvoTypes, verbosity::Integer, fitresult, cache, A, y)
     grow_evotree!(fitresult, cache)
   end
 
-  report = (feature_importances=importance(fitresult, A.names),)
+  # report = (feature_importances=importance(fitresult, A.names),)
```
jeremiedb, line 49: Remove commentted out line.

src/MLJ.jl
```diff
@@ -70,6 +72,22 @@ function predict(::EvoTreeGaussian, fitresult, A)
...
+
+# # Feature Importances
+MMI.reports_feature_importances(::Type{<:EvoTreeClassifier}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeRegressor}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeCount}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeGaussian}) = true
+
+function MMI.feature_importances(m::Union{EvoTreeClassifier, EvoTreeRegressor, EvoTreeCount, EvoTreeGaussian}, fitresult, report)
+    fi_pairs = importance(fitresult, report.features)
+    sort!(fi_pairs, by= x->-x[2])
```
jeremiedb, line 84: Can be removed since `importance` already sort the feature in decreasing order.

src/MLJ.jl
```diff
@@ -70,6 +72,22 @@ function predict(::EvoTreeGaussian, fitresult, A)
   return [Distributions.Normal(pred[i, 1], pred[i, 2]) for i = 1:size(pred, 1)]
 end
 
+
+# # Feature Importances
+MMI.reports_feature_importances(::Type{<:EvoTreeClassifier}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeRegressor}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeCount}) = true
+MMI.reports_feature_importances(::Type{<:EvoTreeGaussian}) = true
```
jeremiedb, line 80: Can be simplified by using `EvoTypes`: 
```julia
MMI.reports_feature_importances(::Type{<:EvoTypes}) = true
```

jeremiedb (changes requested):
Thanks for bringin this PR! 
I've made some comments regarding suggested edits.

john-waczak:
@jeremiedb Comments should be gone now, update method is fixed, and I removed the repeated calls to `reports_feature_importances` per your suggestion. Please let me know if there's anything else I need to  tweak!

jeremiedb:
@john-waczak Thans for the update. Just for clarification, although I understand that MLJ provides a seperate `features importance` extractor function, was there actually a concern for having the feature importance calculated within the report? Is the reason to save compute time while iterating over a model?

OkonSamuel:
> @john-waczak Thans for the update. Just for clarification, although I understand that MLJ provides a seperate `features importance` extractor function, was there actually a concern for having the feature importance calculated within the report? Is the reason to save compute time while iterating over a model?

@jeremiedb MLJ intends to use that API to build other functionality such are recursive feature elimination etc. But it is still going to be accessible by the user when calling `feature_importances(mach::Machine)`. There is nothing stopping you from leaving feature importances in your report then just returning that again in when one calls `feature_importances(model, fitresult, report)`. 
Again, pushing the feature importances calculations to the `feature_importances` method can help to save some computation time. It is only computed when the user needs it.

john-waczak:
@jeremiedb @OkonSamuel I can add them back to the report if you prefer. I just figured with the newer extraction method it made sense to remove the duplicate in the report.

jeremiedb (approved):

jeremiedb:
I just wanted to make sure I understood the original motivation for the changes. Sounds good to me, thanks for @john-waczak and @OkonSamuel for the clarifications.

ablaom:
Just wanted to be sure that the importances returned has the correct form. Is an `AbstractVector` ([required](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Feature-importances)) or `Tuple` of pairs returned? In any case, I suggest explicitly testing the importances have the required form.

ablaom:
Okay, I didn't update my page and now see this is merged with a review from Samuel, so all should be good.

PR #168, opened 2022-10-04
Offset

jeremiedb:
- Breaking change on `fit_evotree` args.


jeremiedb:
To do:

- [ ] Fix eval functions naming (ex `poisson_deviance`). 
- [ ] Align metric tracking with EvoLinear

PR #177, opened 2022-10-26
update oudated doc string

Moelf:


jeremiedb:
Closng in favor of#181 with broader coverage of the docs that needed a refresh. Thanks for raising this!

PR #180, opened 2022-10-27
Fix MLJ metadata declarations so that new docstrings get written to the MLJ model registry

ablaom:
Currently:

```julia
julia> using MLJModels

julia> doc("EvoTreeClassifier")
  Multi-classification with softmax and cross-entropy loss.
```

This is also an issue for EvoLinear.jl and the fix required is likely the same. 

Follow up required:

- [ ] Update the MLJ model registry

### DIFF
--- a/src/MLJ.jl
+++ b/src/MLJ.jl
@@ -94,11 +94,6 @@ end
 
 
 # Metadata
-const EvoTreeRegressor_desc = "Regression models with various underlying methods: least square, quantile, logistic, gamma (deviance), tweedie (deviance)."
-const EvoTreeClassifier_desc = "Multi-classification with softmax and cross-entropy loss."
-const EvoTreeCount_desc = "Poisson regression fitting λ with deviance minimization."
-const EvoTreeGaussian_desc = "Deprecated - Use EvoTreeMLE with `loss=:normal` instead. Gaussian maximum likelihood of μ and σ."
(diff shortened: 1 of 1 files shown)


jeremiedb:
Sorry about the naive question, but I'm not sure to understand what's the issue with the example provided: 
```julia
julia> doc("EvoTreeClassifier")
  Multi-classification with softmax and cross-entropy loss.
```

What's the expected outcome?

ablaom:
You should get the full doc-string: the same as if you do "using EvoTrees.jl; @doc EvoTreeClassifier".

The MLJ model trait `docstring` (alias `descs`) used to be for a short summary string, which was not that useful, in retrospect. Now it is not to be overloaded but instead falls back to full docstring. More at https://github.com/alan-turing-institute/MLJ.jl/issues/898 .

src/MLJ.jl
```diff
@@ -119,7 +119,6 @@ MMI.metadata_model(
     target_scitype = AbstractVector{<:MMI.Continuous},
     weights = false,
     path = "EvoTrees.EvoTreeRegressor",
-    descr = EvoTreeRegressor_desc,
```
jeremiedb, line 122: I guess we can remove the short descriptions in lines 97-101 as well?

ablaom: Good catch. Done.

PR #183, opened 2022-11-02
Logger

jeremiedb:
Refactor of the fitting APIs. Can now return eval tracking info. 
Breaking for classification, reqwuires `num_class` in constructor. 
No more export of `grow_tree!` and  `predict`. For the latter, recommend to use `(m::EvoTree)(x)`.

### DIFF
--- a/src/MLJ.jl
+++ b/src/MLJ.jl
@@ -5,24 +5,16 @@ function MMI.fit(model::EvoTypes, verbosity::Int, A, y, w = nothing)
     else
         fitresult, cache = init_evotree(model; x_train = A.matrix, y_train = y, w_train = w)
     end
-    grow_evotree!(fitresult, cache)
+    while cache[:info][:nrounds] < model.nrounds
+        grow_evotree!(fitresult, cache, model)
+    end
     report = (features = A.names,)
     return fitresult, cache, report
 end
 
-function okay_to_continue(new, old)
-    new.nrounds - old.nrounds >= 0 &&
-        new.lambda == old.lambda &&
-        new.gamma == old.gamma &&
-        new.max_depth == old.max_depth &&
-        new.min_weight == old.min_weight &&
-        new.rowsample == old.rowsample &&
-        new.colsample == old.colsample &&
-        new.nbins == old.nbins &&
-        new.alpha == old.alpha &&
-        new.device == old.device
+function okay_to_continue(model, fitresult, cache)
+    model.nrounds - cache[:info][:nrounds] >= 0
 end
-
 
 # Generate names to be used by feature_importances in the report
 MMI.reformat(::EvoTypes, X, y, w) =
@@ -55,8 +47,10 @@ function MMI.update(
     y,
     w = nothing,
 )
-    if okay_to_continue(model, cache.params)
-        grow_evotree!(fitresult, cache)
+    if okay_to_continue(model, fitresult, cache)
+        while cache[:info][:nrounds] < model.nrounds
+            grow_evotree!(fitresult, cache, model)
+        end
     else
         if model.device == "gpu"
             fitresult, cache =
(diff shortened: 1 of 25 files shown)


src/MLJ.jl
```diff
@@ -5,25 +5,18 @@ function MMI.fit(model::EvoTypes, verbosity::Int, A, y, w = nothing)
...
-        new.max_depth == old.max_depth &&
-        new.min_weight == old.min_weight &&
-        new.rowsample == old.rowsample &&
-        new.colsample == old.colsample &&
-        new.nbins == old.nbins &&
-        new.alpha == old.alpha &&
-        new.device == old.device
+function okay_to_continue(model, fitresult, cache)
+    return model.nrounds - cache[:info][:nrounds] >= 0 &&
+           all(get_types(model) .== get_types(fitresult))
```
ablaom, line 17: You sure this is correct? Does `get_types` exclude `:nrounds`, then?

You can also use `MLJModelInterface.is_equal_except(new, old, :nrounds)`.

ablaom:
> Breaking for classification, requires num_class in constructor.

Is this true of the MLJ interface model as well? That would make these models the only (non-wrapper) models with a compulsory hyperparameter. Can't you get the classes from the class pool of `y`, which is necessarily a categorical vector? The pool has all the classes, even if they are not explicit elements of `y`, right? I mean, this is our raison d'etre for using categorical vectors: [robustness of pool to subsampling](https://alan-turing-institute.github.io/MLJ.jl/dev/working_with_categorical_data/#Tracking-all-levels).

src/MLJ.jl
jeremiedb: Yes, at least it's aligned with that I intended to do, that is to relax the check so that other training configuration can change throughout training, for example the eta, or some other regularization term. The `get_types` enforce the check on the loss type, the floating point, and the output size.

jeremiedb:
Yes the `num_class` would become mandatory for the `EvoTreeClassifier`, so it would impact as well MMI. 
I thought it wouldn't be much of a concern given it borrows from XGboost requirement. The change allowed to include the output dimensio within the type, resulting in neater code and may prove helpful in for some performance gains. 
I can double check if this `num_class` could be avoided from the constructor and assess the impacts of reverting on this change.

jeremiedb:
Although there could be a way aroud it by adding arguments to a couple of functions, I'd say that the `num_class` argument is really helpful in the code cleanup I've bee trying into to facilitate the ongoing maintenance and much of the complications I unnecessarily introduced especially in the early days. 
Apart from not being aligned with other classification methods, are there other practical implications for the user having to supply the numbe of classes? I'm assuming it shouldn't be much of a assle considereing the information should be easily available at the time of callnig the model constructor.

ablaom:
One issue is that the same model instance will not work for different datasets. For example, I could imagine MLJ providing a benchmarking tool, where we are testing a given classifier on *multiple* datasets (with different numbers of target classes) and EvoTreeClassifier is not going to fit into such a scheme, because the hyper-parameters have to be adjusted to the dataset. I think that as far as possible we try to separate properties of the data from model hyper-parameters.  I'd suggest the extra complexity in your case is a reasonable price to pay for that. 

We're already running into a problem at [MLJTestIntegration.jl](https://github.com/JuliaAI/MLJTestIntegration.jl) because there we already assume MLJ models provide a default instance, ie, `MyModel()` is expected to work. So we've had to remove EvoTreeClassifier from our global integration tests.

jeremiedb:
@ablaom you're making my life hard :P 
I've reverted the changes to remove the need for `num_class` for Classifier. It's up to date on `main` branch. 
Could you confirm it fixes compatibility with [MLJTestIntegration.jl](https://github.com/JuliaAI/MLJTestIntegration.jl)?

ablaom:
Yes, all good: https://github.com/Evovest/EvoTrees.jl/pull/191

PR #186, opened 2022-11-02
massively reduce CPU allocation

Moelf:
setup:
```julia
julia> x_train = rand(10^4, 60);

julia> y_train = rand(size(x_train, 1));

julia> config = EvoTreeRegressor(
               loss=:logistic,
               nrounds=100,
               lambda=0.5, gamma=0.0, eta=0.05,
               max_depth=6, min_weight=1.0,
               rowsample=1.0, colsample=0.5, nbins=64);
[ Info: Following 5 arguments were not provided and will be set to default: alpha, T, rng, device, monotone_constraints.
```

## Before
```julia
julia> @time fit_evotree(config; x_train, y_train);
  0.330572 seconds (440.90 k allocations: 218.696 MiB, 9.79% gc time)
```

## After this PR
```julia
julia> @time fit_evotree(config; x_train, y_train);
  0.242588 seconds (343.28 k allocations: 54.385 MiB, 6.30% gc time)
```

### DIFF
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -161,11 +161,13 @@ function grow_tree!(
 ) where {L,K,T}
 
     # reset nodes
-    @threads for n in eachindex(nodes)
-        [nodes[n].h[j] .= 0 for j in 𝑗]
-        nodes[n].∑ .= 0
-        nodes[n].gain = 0
-        fill!(nodes[n].gains, -Inf)
+    @threads for n in nodes
+        for j in 𝑗
+            n.h[j] .= 0
+        end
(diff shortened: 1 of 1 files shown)


Moelf:
@jeremiedb btw, I think for all the `Vector{Vector{}}` stuff, you should use https://github.com/JuliaArrays/ArraysOfArrays.jl

its supports syntax of `Vector{Vector{}}` while also providing flattenview which can help you quickly updates all the underlying values in node (without pointer chasing) and also GPU friendly.

PR #191, opened 2022-11-08
Add generic MLJ interface tests

ablaom:
This PR is only a suggestion, as MLJTestIntegration is at an experimental stage. We do have this installed in a couple of other packages. 

 Adding MLJTestIntegration currently adds the entire MLJ stack (MLJTuning, MLJIteration, etc), so it's a bigger (test) dependency than what is currently here. But I expect we will factor out a lighter-weight MLJTestInterface in the near future (and the syntax won't change much, if at all).

The `level=2` tests included here are: 

    •  :model_type: Load model type using `load_path(model_type)` (to check `load_path` trait
       is correctly overloaded).

    •  :model_instance: Create a default instance.

    •  :fitted_machine: Bind instance to data in a machine and fit!. Call report and
       fitted_params on the machine.

    •  :operations: Call implemented operations, such as predict and transform

But I have independently checked that level=4 tests also pass. Do `using MLJTestIntegration; @doc MLJTestIntegration.test` for further details.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
-authors = ["jeremiedb <jeremie.db@evovest.com>"]
 name = "EvoTrees"
 uuid = "f6006082-12f8-11e9-0c9c-0d5d367ab1e5"
+authors = ["jeremiedb <jeremie.db@evovest.com>"]
 version = "0.13.0"
 
 [deps]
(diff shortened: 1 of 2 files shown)


jeremiedb:
Looks good, thanks for your work on MLJ integration!

PR #200, opened 2022-12-14
Update benchmarks-regressor.jl

rwarnung:
Parameter max_depth was set to 5 for evoTrees for consistency with xgboost.


jeremiedb:
The difference is on purpose as the interpretation is different in EvoTrees than XGBoost. For EvoTrees, depth of 1 would be a single leaf, for which I'm not sure there's an equivalent in XGBoost (is depth == 0 supported?). Perhaps that can be understood from Julia's 1-based index :) Though I understand this may leads to some confusion for people working with other boosted trees implementations.

rwarnung:
@jeremiedb Oh my mistake. I didn't know that. I will read the docs more clearly :) So this PR doesn't make any sense. What I still have to figure out is the speed improvement. When I use max.depth+1 for evoTress and compare it to max.depth for xgboost, then I don't see the speed. I have to perform more experiments, it seems. Thank you!

jeremiedb:
If you get same training speed for both EvoTrees and XGBoost, it would be useful if you could provide the data dims (nobs, nfeats), hyper-params setup as well as hardware on which it's being executed (notably physical cores and total threads)

rwarnung:
Hi @jeremiedb ,
thank you for your comments and taking the time to discuss this with me.

The summary of my further experiments is:
1. I do see the 50% speed-up in Julia.
2. Calling xgboost ("directly") from R and calling `evoTrees `via `JuliaCall`, evoTrees is 50% slower than xgboost.

Of cours, point 1 supports the speed of the package. This is great! Point 2 points to some overhead when usign `JuliaCall`.

Nevertheless, I don't yet give up on point 2 as this would be nice use case (calling julia from R for specific tasks).

Do you have similar experiences with calling julia from R? 
Thank you and kind regards,
Richard

------------

PS. The details:

First I tried to replicate the speed comparison using basically your file.
I did immaterial adaptions here: https://github.com/rwarnung/testing_evotrees/blob/main/speed_comparison.jl

Basically, it is the set-up with 100 features, 1 mio observations and 200 rounds. The data is random.
I can use only one thread. My local machine is an ultra book (I will later do experiments on AWS) with Intel Core i7-8550U CPU @ 1.80 Ghz.
It has 4 kernels and 8 logical processors. When I determine the number of threads in Julia, then I can only use 1. I am not a hardware guy, I hope this makes sense :)

xgboost train took 117 seconds.
xgboost predict took 4 seconds.

pars were

params_xgb = Dict(
    :num_round => nrounds,
    :max_depth => 5,
    :eta => 0.05,
    :objective => loss_xgb,
    :print_every_n => 50,
    :subsample => 0.5,
    :colsample_bytree => 0.5,
    :tree_method => "hist",
    :max_bin => 64,
)


evoTrees train took 61-63 seconds (tried it various times).
evoTrees predict took 4 seconds.

pars were

params_evo = EvoTreeRegressor(
    T=T,
    loss=loss_evo,
    nrounds=nrounds,
    alpha=0.5,
    lambda=0.0,
    gamma=0.0,
    eta=0.05,
    max_depth=6,
    min_weight=1.0,
    rowsample=0.5,
    colsample=0.5,
    nbins=64,
    rng = 123,
)

package versions 
⌃ [69666777] Arrow v2.3.0
  [6e4b80f9] BenchmarkTools v1.3.2
  [b99e7846] BinaryProvider v0.5.10
⌃ [336ed68f] CSV v0.10.7
  [324d7699] CategoricalArrays v0.10.7
⌃ [a93c6f00] DataFrames v1.4.2
  [f6006082] EvoTrees v0.14.2
  [7acf609c] LightGBM v0.6.0
  [46a55296] ParquetFiles v0.2.0
  [6f49c342] RCall v0.13.14
  [295af30f] Revise v3.4.0
  [2913bbd2] StatsBase v0.33.21
  [fd094767] Suppressor v0.2.1
  [c4f8c510] UMAP v0.1.9
  [856ac37a] UrlDownload v1.0.1
⌃ [009559a3] XGBoost v2.0.0

--------------------------------------
Comparison with R:
https://github.com/rwarnung/testing_evotrees/blob/main/speed_comparison.R

##### xgb.Booster
raw: 823.2 Kb 
call:
  xgb.train(params = params_xgb, data = dtrain, nrounds = nrounds, 
    verbose = 1, print_every_n = 50L)
params (as set within xgb.train):
  booster = "gbtree", eta = "0.05", max_depth = "5", min_child_weight = "1", subsample = "0.5", colsample_bytree = "0.5", gamma = "0", tree_method = "hist", objective = "reg:squarederror", max_bin = "64", validate_parameters = "TRUE"
xgb.attributes:
  niter
callbacks:
  cb.print.evaluation(period = print_every_n)
niter: 200
nfeatures : 100 

------------

##
EvoTreeRegressor{EvoTrees.Linear, Float32}(200, 0.0f0, 0.0f0, 0.05f0, 6, 1.0f0, 0.5f0, 0.5f0, 64, 0.5f0, Dict{Int64, Int64}(), Random.TaskLocalRNG(), "cpu")

------------

100 features, 200 rounds, 500 000 obs
xgboost: 
train: 4 sec

evotrees: 
train: 28 seconds

cor between predictions: 0.47

----------
100 features, 200 rounds, 1000 000 obs
xgboost: 
train: 40-41 sec

evotrees: 
train: 60-61 seconds

cor between predictions: 0.38

------------
> sessionInfo()
R version 4.1.1 (2021-08-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 22621)

Matrix products: default

locale:
[1] LC_COLLATE=German_Austria.1252  LC_CTYPE=C                      LC_MONETARY=German_Austria.1252 LC_NUMERIC=C                   
[5] LC_TIME=German_Austria.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] xgboost_1.5.0.2  JuliaCall_0.17.5

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.7        rstudioapi_0.13   knitr_1.40        feather_0.3.5     magrittr_2.0.1    hms_1.1.1         bit_4.0.4        
 [8] lattice_0.20-44   R6_2.5.1          rlang_1.0.6       fansi_0.5.0       tools_4.1.1       tictoc_1.0.1      grid_4.1.1       
[15] data.table_1.14.2 xfun_0.34         pacman_0.5.1      coro_1.0.2        utf8_1.2.2        cli_3.4.1         ellipsis_0.3.2   
[22] bit64_4.0.5       tibble_3.1.8      lifecycle_1.0.3   Matrix_1.3-4      processx_3.5.2    torch_0.6.0       readr_2.1.1      
[29] callr_3.7.0       tzdb_0.2.0        vctrs_0.5.0       ps_1.6.0          glue_1.5.1        compiler_4.1.1    pillar_1.8.1     
[36] jsonlite_1.7.2    pkgconfig_2.0.3

jeremiedb:
It effectively appears like the communication overhead between RCall and Julia adds a significant burden. 
One possible alternative to test would be `JuliaCOnnector`:  https://github.com/stefan-m-lenz/JuliaConnectoR. I've never used but it could be worth testing if its communication method is more efficient. 

If using EvoTree with Julia call, in case you might have missed, there's a dedicated R package as well: https://github.com/Evovest/EvoTrees. I haven't updated for some time so it will likely hit some issue, but it may give insight on a Julia-R integration. 

Finally, regarding the number of threads, you likely want to have Julia to use multiple threads. In case your Julia launches by default with 1 thread, then you could consider adding `JULIA_NUM_THREADS` as an environment variable and set it to your number of cores. When in Julia, this can be confirmed with : `Base.Threads.nthreads()`.

rwarnung:
Thank you for all these valuable comments. I will consider them and come
back to you, if this is also of interest for you.
Kind regards
Richard

jeremiedb ***@***.***> schrieb am Fr., 23. Dez. 2022, 06:52:

jeremiedb:
@rwarnung 

For info, I did test the R / Julia interface of JuliaConnectoR and it seems to offer an improved performance, at least for training times: 

```R
library(ggplot2)
library(xgboost)
library(JuliaConnectoR)

nrows <- 1e6
ncols = 100
x <- matrix(runif(nrows * ncols), nrow = nrows)
y <- runif(nrows)
target_train <- y
data_train <- x

EvoTrees <- JuliaConnectoR::juliaImport("EvoTrees", all = F)
regressor <- EvoTrees$EvoTreeRegressor(nrounds = 200, eta = 0.05, rowsample = 0.5, colsample = 0.5, nbins = 64)
system.time(model_evo <- EvoTrees$fit_evotree(regressor, x_train = data_train, y_train = target_train))
   user  system elapsed 
   0.24    0.59   11.93 
```

Sad part is that inference takes more time than the training! It suspects this is due to the predictions being directly transferred back to R, and perhaps that steps is inefficient as it is (although from the fitting step, R to Julia data transfers appears quite effective as there doesn't seem to be any overhead in fitting time compare to pure Julia.  

```R
system.time(pred_evo <- EvoTrees$predict_evotree(model_evo, data_train))
   user  system elapsed 
   0.23    0.26   18.50

rwarnung:
@jeremiedb thank you very much for coming back to me! I had a look at JuliaConnectoR and unfortunately I didn't know how to call the predict function. If I run your code (with the most recent versions of EvoTrees and JuliaConnectoR) then I get the following error message:

> Error in system.time(pred_evo <- EvoTrees$predict_evotree(model_evo, data_train)) : 
>   attempt to apply non-function
> 

It is confusing for me that things work on your end which do not work on mine. What I also had to figure out (at least I thought) was to set the type right. As input data from R are Float64 in Julia I set the parameter  `T ` for EvoTrees (in my analyses, not in the attempt to replicate yours) by

`EvoTreeRegressor(T = juliaExpr("Float64"))`.

This came from a [discussion with the author of JuliaConnectoR](https://stackoverflow.com/questions/74977087/how-to-pass-a-julia-type-in-the-r-package-juliaconnector/74983668#74983668).

This seems not to be required in your code ... I am confused :)

I am somehow stuck with JuliaConnectoR, that's why I focus on JuliaCall. I am still in the process of evaluating comparisons of EvoTrees with lightgbm and xgboost. So far it feels that EvoTrees is most similar to lightgbm in terms of correlation of predictions and run time.

jeremiedb:
Sorry I forgot to mention a very relevant point: this above was tested from the `juliaConnectoR` branch:  https://github.com/Evovest/EvoTrees.jl/tree/juliaConnectoR
So you'd need to to have EvoTrees installed with something like: `] add EvoTrees#juliaConnectoR`
The modification I made was to add a `predict_evotree` and have that function exported. It's really just `predict_evotree = predict`, but I wouldn't want to export `predict` as it's more susceptible to cause naming ambiguities with other packages.

jeremiedb:
Otherwise, the `T` argument shouldn't be an issue as this argument is meant to refer to whether the tree building algo will be execute on Float32 or Float64, while `fit_evotree` should be equally happy handling Float32 or Float64 Matrix. The downside being that you will have some extra memory allocation for the one time conversion between Float32 <-> Float64 is your inputs doesn't match the selected `T`. For regular regressin, You'll get some meaningful speedup working on the default `T = Float32`. 

So, not sure about the origin of the type error you experienced. Maybe first test if the above JuliaConnectoR works on your end using the `juliaConnectoR` branch?

rwarnung:
@jeremiedb thank you for your help! 
- with the correct branche I can use `EvoTrees `with `juliaConnectoR `. I even see a 40% speed-up as compared to xgboost!
- yes, I don't have to set the parameter `T`. It was not the reason for previous errors.

Thank you! Now I can use `EvoTrees `in R in two ways (with JuliaCall and juliaConnectoR) :) (the install of the R-package for `EvoTrees `did not work the last time I tried it. But with those two packages this is not that relevant).

Thank you once again for your work on this package and for discussing it with me!

Issue #216, opened 2023-03-23
MethodError when using EvoTreeClassifier with MLJTuningTreeParzen


svilupp:
Behavior: When running the [MLJ tuning example](https://github.com/IQVIA-ML/TreeParzen.jl/blob/master/docs/examples/simple_mlj_demo/simple_mlj_demo.md) with EvoTreeClassifier, tuning fails due to a MethodError that ‘ EvoTreeClassifier{Softmax,Float32}(;…)’ doesn’t exist.

Root cause: I believe the issue comes from [this line](https://github.com/IQVIA-ML/TreeParzen.jl/blob/15224294dce354ce2148bea4e37066c2afa44654/src/MLJTreeParzen.jl#L266), which new hyperparams to a kwarg constructor extracted as ‘typeof(model)(; new_params…)’

Unfortunately, that requires a constructor ‘ EvoTreeClassifier{Softmax,Float32}(;…)’, but we have only the [generic one defined](https://github.com/Evovest/EvoTrees.jl/blob/2ff0f5a7a1ab7e85326212cf9c9bd48ead43cc50/src/models.jl#L194).

It can be solved by adding new kwarg constructors for more specific parametric types (eg, with ‘where {L,T}’).

Impact: I believe this impacts all Evo models when used with MLJTuningTreeParzen.

Proposal: I’m out until next week, but I’m happy to open the PR when I’m back.

For completeness, this happens on the latest version of EvoTrees. Julia 1.8.5., arm-based.

jeremiedb:
Thanks for reporting this. If you can open the PR, that would be welcomed!

ablaom:
I think the indicated line in TreeParzen.jl is quite brittle. I can imaging it failing for other MLJ models that don't have parameter-type-specific constructors. 

I'll open an issue at TreeParzen.jl. Unfortunately, this pkg has only sporadic maintenance. So you might want to go ahead with the local fix, which sounds innocuous enough.

ablaom:
@svilupp Thanks for reporting, by the way.

PR #221, opened 2023-04-21
add kwarg constructor

svilupp:
Fixes https://github.com/Evovest/EvoTrees.jl/issues/216

This PR should be ready for review.

Until the issue with parametric kwargs is fixed upstream in MLJTreeParzen, these constructors will prevent hard-to-debug errors for new users.

Changes:
- added kwarg constructor for parametric types, eg, `EvoTreeRegressor{Linear,Float32}(; kwargs...)`
- added related tests
- I did a manual formatting clean-up to match the style of the rest of the code base (`JuliaFormatter.toml` is missing)

Tests:
- Core tests pass
- MLJ integration tests fail on an unrelated issue -- Should I open a separate issue? I think we're missing some compat bounds (more below)

Test failure (unrelated to the PR):
- I get test failure outside of @test, `UndefVarError: make_binary not defined` (same for make_count)
- It's caused by my test environment using MLJTestIntegration v0.1.0 but these functions were introduced only later
- I believe a compat entry should be added for at least 0.2.0 or above


svilupp:
Looking at the failing tests, it all seems to be related to the missing compat for MLJTestIntegration mentioned above. 
I've added a compat for 0.4 to see if it helps.

jeremiedb:
I wouldn't want to add `MLJTestIntegration` as a dependency to the package per se, as I'm loooking to keep the dependencies as light as possible. 

An approach could be to go for a `Project.toml` in the `test/` directory, so as to add a specific compat, instead of the `extras` approach used so far. 

However, I quite don't understand why `MLJTestIntegration` resolves to v0.1 while it has not been an issue up to now. Tests also work fine when I run them locally, so I'm a little worried there's something unexpected that has changed. Any idea on that?

svilupp:
> However, I quite don't understand why MLJTestIntegration resolves to v0.1 while it has not been an issue up to now.  

I’ll look into it. 

I assumed it was just cached somewhere, but I’ll check the dep graphs. It’s unlikely to be caused by this PR, as I haven't touched any of the failing bits or imports, so maybe something upstream?

> I wouldn't want to add MLJTestIntegration as a dependency to the package per se, as I'm loooking to keep the dependencies as light as possible.

Could you please clarify this? I didn’t change the deps organization (it was already in extras for testing the MLJ integration), so nothing was added?

> An approach could be to go for a Project.toml in the test/ directory, so as to add a specific compat, instead of the extras approach used so far.

I thought this is equivalent. Is it not?
Ie, it’s installed only when testing, but the difference is that if you have dedicated Test Env, you’ll have to update your compats in two places

svilupp:
So I've used TestEnv to explore the test environment on main branch. 
The same issue happens -> tests fail because of outdated MLJTestIntegration (0.1.0)

This is the status of `]st -m --outdated`
```
⌅ [30fc2ffe] LossFunctions v0.8.1 (<v0.9.0): MLJBase
⌅ [add582a8] MLJ v0.18.6 (<v0.19.1): MLJTestIntegration
⌅ [a7f614a8] MLJBase v0.20.20 (<v0.21.9): MLJ
⌅ [d491faf4] MLJModels v0.15.15 (<v0.16.6): MLJ
⌃ [697918b4] MLJTestIntegration v0.1.0 (<v0.4.0)
⌅ [458c3c95] OpenSSL_jll v1.1.20+0 (<v3.0.8+0): OpenSSL
```
There is nothing blocking the update (and the other MLJ deps are held back by MLJTestIntegration).

After running the update, all tests pass.

If I delete my Manifest.toml and instantiate again, the error appears again (because the environment automatically picks 0.1.0 -- can't tell why).

svilupp:
> An approach could be to go for a Project.toml in the test/ directory, so as to add a specific compat, instead of the extras approach used so far.

I've extracted the test environment to a dedicated `test/Project.toml` as requested + updated the name of this PR to reflect the change in scope.

It fixes the testing failures on my machine. Let's see what the CI thinks.

Given the cleaning up nature, would you consider adding a JuliaFormatter.toml spec? (I'd do it in a separate PR not to mix too many things)

jeremiedb:
Thanks for looking further into this. My understanding is that there must have been some recent changes in the MLJ's dependency stack that result in those conflicts. I think I'd prefer seeing this resolve on that end rather change the Test dependency structure. 

@ablaom The versioning conflict experience in this PR seems rooted from `MLJTestIntegration.jl` having a dependency on `NearestNeighborModels` which itself has a compat bound that doesn't include `FillArrays` recent v1 release. A patch release to cover v1 on  `NearestNeighborModels` would fix this, though it I'm not sure whether it's intended to have `NearestNeighborModels` as a dependency in `MLJTestIntegration` as I expect such package not to depend on specific model implementation. Let me know what you think!

ablaom:
You should not be using `MLJTestIntegration.jl`, as is loads the whole MLJ shebang.

Use `MLJTestInterface.jl` instead. See, for example, here: https://github.com/JuliaAI/MLJXGBoostInterface.jl/blob/4456438d134c20207c0c8383afb14d9f7bfa816c/test/runtests.jl#L205

svilupp:
I've replaced MLJTestIntegration with MLJTestIntegration, all tests pass locally.

@jeremiedb, 
- I've kept the separate test environment, as per your request on minimizing the dependencies for EvoTrees
- I've added a compat for MLJTestInterface to avoid unrelated test failures in the future (like what we had here)

jeremiedb:
@svilupp Sorry for this apparent back and forth, but given the issue was from usage of `MLJTestIntegration` instead of `MLJTestInterface`, I'd rather move back to the original approach that use `[extras]` and  `[targets]` in the main `Project.toml`. This would keep the Project in line with XGBoost implementation and the approach still used by most packages. 

@ablaom Thanks for the clarification!

svilupp:
@jeremiedb No problem. I've moved it back and all tests pass.

PR should be ready for review.

jeremiedb:
@svilupp Looks good to go, thanks!

Issue #222, opened 2023-04-21
InexactError: trunc(UInt8, 256) when user sets nbins>255


svilupp:
Current behaviour: If a user sets `nbins>255` they will get `InexactError: trunc(UInt8, 256)`

Expected behaviour: Either the restriction should be explicit and caught on initialization or the code should be updated.

Possible cause:
- Once features are split, all data points are bucketed into corresponding slots by function `binarize`
- This [line](https://github.com/Evovest/EvoTrees.jl/blob/f95e1b65612c6413f3d2f770469a9f74019d890a/src/find_split.jl#L21) makes an assumption that the buckets are of type `UInt8`, so implicitly assuming there are never more than 255 bins

Suggested solution:
- 255 bins should be sufficient and no user has run into this limitation so far, so it can be kept
- we should add a note in the documentation
- and check this value on initialization (with a helpful error message)

I'm happy to take a stab at the PR if helpful.

svilupp:
MWE (taken from the example in docs):
```
using EvoTrees
using Random

x_train = randn(100, 10)
y_train = randn(100)

config = EvoTreeRegressor(
    loss=:linear, 
    nrounds=100, 
    max_depth=6, 
    nbins=256,
    eta=0.1,
    lambda=0.1, 
    gamma=0.1, 
    min_weight=1.0,
    rowsample=0.5, 
    colsample=0.8)

m = fit_evotree(config; x_train, y_train)
preds = m(x_train)

# throws: nested task error: InexactError: trunc(UInt8, 256)
```

Tested on the latest version in main. 
versioninfo()
> Julia Version 1.8.5
> Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
> Platform Info:
>   OS: macOS (arm64-apple-darwin21.5.0)
>   CPU: 8 × Apple M1 Pro
>   WORD_SIZE: 64
>   LIBM: libopenlibm
>   LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
>   Threads: 6 on 6 virtual cores

jeremiedb:
Adding domain validation of the arguments passed to the hyper-params constructors is something I've been shoveling forward a bit too long! Thanks for reporting, feel free to go for a PR, otherwise, I'll likely get it done with 1-2 days.

svilupp:
I have too many other things going on this weekend, but I should be able to get to it on Monday. 

When I find time, I'll check here if you've started or not.

[...]

PR #223, opened 2023-04-23
Type validation for model parameters

svilupp:
Fixes https://github.com/Evovest/EvoTrees.jl/issues/222

An initial proposal to validate whether this design is acceptable.
- Restricts the following parameters: `nbins` (to UInt8)
- Adds a function to convert safely (`convert_parameter`) with an understandable error for the user (see example below)

MWE:
```
config = EvoTreeClassifier(nbins=255)
# works fine

config = EvoTreeClassifier(nbins=256)
# ERROR: Invalid value for parameter `nbins`: 256. `nbins` must be of type UInt8 (ie, value between 0 and 255).
```

Question:
- Are there more parameters that should be more restrictive? So far I've restricted only `nbins` to UInt8

TODO:
- corresponding tests once we agree on the approach

### DIFF
--- a/docs/src/tutorials/classification-iris.md
+++ b/docs/src/tutorials/classification-iris.md
@@ -1,4 +1,4 @@
-# Classication on Iris dataset
+# Classification on Iris dataset
 
 We will use the iris dataset, which is included in the MLDatasets package. This dataset consists of measurements of the sepal length, sepal width, petal length, and petal width for three different types of iris flowers: Setosa, Versicolor, and Virginica.
 
--- a/src/MLJ.jl
+++ b/src/MLJ.jl
@@ -193,7 +193,7 @@ A model type for constructing a EvoTreeRegressor, based on [EvoTrees.jl](https:/
 - `max_depth=5`:          Maximum depth of a tree. Must be >= 1. A tree of depth 1 is made of a single prediction leaf.
   A complete tree of depth N contains `2^(N - 1)` terminal leaves and `2^(N - 1) - 1` split nodes.
   Compute cost is proportional to `2^max_depth`. Typical optimal values are in the 3 to 9 range.
-- `min_weight=0.0`:       Minimum weight needed in a node to perform a split. Matches the number of observations by default or the sum of weights as provided by the `weights` vector.
+- `min_weight=0.0`:       Minimum weight needed in a node to perform a split. Matches the number of observations by default or the sum of weights as provided by the `weights` vector. Must be > 0.
 - `rowsample=1.0`:        Proportion of rows that are sampled at each iteration to build the tree. Should be in `]0, 1]`.
 - `colsample=1.0`:        Proportion of columns / features that are sampled at each iteration to build the tree. Should be in `]0, 1]`.
 - `nbins=32`:             Number of bins into which each feature is quantized. Buckets are defined based on quantiles, hence resulting in equal weight bins. Should be between 2 and 255.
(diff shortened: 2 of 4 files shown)


jeremiedb:
I like the detailed error message. 
Chaging the type of the the `struct` such as `EvoTreeRegressor` would however in a braking change. As such, I'd favor instead a mechanism based on the parameter domain rather their their type. For example, let the `nbins` as a Int, but rather check that it's value is [2-255]. 

The other checks would include:
- nrounds >= 0
- eta >= 0
- lambda >= 0
- gamma >= 0
- alpha [0-1]
- max_depth >= 1
- min_weight >= 0
- rowsample/colsample ]0, 1]

Let me know if you want to pursue in that direction, or I can follow up on this.

svilupp:
I've updated the implementation only for EvoTreeRegressor to sense-check if that's what you were thinking.

I created a dedicated function to avoid copy-paste errors `check_args` that can be simply called in each kwarg constructor before instantiating the model struct.

Is this closer to what you had in mind?

svilupp:
> Changing the type of the struct such as EvoTreeRegressor would however in a breaking change. 

Just out of curiosity - is it a breaking change if, implicitly, it was always UInt8? 
Ie, any inconsistent number format (that cannot be converted into UInt8) would throw an error.

jeremiedb:
I think the `check_parameter` / `check_args` is a very clean approach! I'm good moving foward with it for all the constructors (Count, Classifier...). As a possible extra, it could be nice to have a `check_args` that acts on the `EvoTreeRegressor` & cie. ni addition to the args `Dict` as in its current form. That would be convenient for applications like in MLJ where the arguments can be mutated so that hyper-params can be reassessed following such mutation. 

Regardng the breaking change of `UInt8`, it would likely be transparent for most users, but I think it should brake a situation where a `EvoTreeRegressor` was saved through serialization. I haven't testetd, but I suspect it would fail to load if the original object was saved as an Int, while the current `struct` is defined as a `UInt8`.

ablaom:
In MLJ parameter checks are implemented by overloaded the `clean!` method. This method is called in MLJ whenever you fit a machine to check users has not mutated the model in a way that is not allowed.

I you use the `@mlj_model` macro (can't remember if this is used here) then that automatically overloads `clean!` for you.

See [here](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#New-model-type-declarations-and-optional-clean!-method), and, for the macro, [here](https://alan-turing-institute.github.io/MLJ.jl/dev/adding_models_for_general_use/#Macro-shortcut).

svilupp:
Thank you for sharing the examples!

Sorry for the long pause, got overwhelmed by work & life :) But I want to close this now.

After having looked through the example of `clean!` method and how it's applied, it seems to be different from our current approach:
- It instantiates the model first
- Then it forces the parameters to their allowed ranges
- It produces a `@warn` log
Basically, as the name suggests, it "cleans it" :)

Whereas `check_args()` checks the type and range before we instantiate the model, and throws an error if it's not acceptable. 
So on one hand, in cases when a model cannot be instantiated, we catch this and provide more helpful error.
On the other hand, the user must fix the error, whereas `clean!` does it under the hood (so the user might not even find out if the logging is disabled)

**Potential Options**
1) keep `check_args()` which will be called on every kwarg construction
2) replace `check_args()` with `MMI.clean!` method inside the kwarg constructors, drop the type check, check only the value range and embrace the `@warn` + "auto-fix" paradigm (minor: would require breaking the separation of MLJ code and model definitions)
3) use `@mlj_model` macro for the structs to define param ranges

I think the last one is out of the question given the discussion in other PRs (backward compatibility).

The choice between 1) and 2) is a bit philosophical. @jeremiedb , do you have a preference?

I personally prefer 1) because it's more explicit and fails on a mistake, but I appreciate that `clean!` is consistent with the rest of MLJ world (although not every model has it). There could also be some hybrids (throw an error instead of auto-fixing etc., it depends on what behaviour we desire)

jeremiedb:
Thanks for looking back into this! 
I'd also favor going with 1 for now. I think it covers the spirit of `clean!` for having sanity check of the hyper params while avoiding taking initiave for a substitute. I prefer treatning improper input values as domain error and as such, returns an error and let the users figure out what's the proper fix for their use case. 

For `eta`, I agree it should be structly positive. 

Regarding: 
> it feels like the models should be validated when constructing directly (via args, not kwargs)

Would you have an example of what you'd have in mind?

Let me know if you need any further clarification.

svilupp:
I've implemented the following changes:
- eta > 0
- nrounds >= 1 (not specified, but it's implicit given the fit_evotree training loop)
- added tests for the checking functionality itself, checking upon construction and checking a model (eg, after a mutation)
- Checking after mutation is done only manually, ie, I have not overloaded the setter
- Updated docs with the implemented parameter value limits
- fixed a few typos that I spotted

Tests pass locally. PR should be ready for review.

jeremiedb:
This looks good to go now, I'll push a patch release. Thanks for your contribution!

PR #259, opened 2023-10-05
Create CUDA extension

### LINKED ISSUE #226: Make CUDA a weak dependency
I think it would be very useful for downstream packages that are mainly/only using the CPU functionality if the CUDA dependency would be made a weak dependency on Julia >= 1.9. That is, in newer Julia versions CUDA and all its dependencies would not be installed and loaded automatically but the GPU features would be loaded only if the user (or some other package loaded by the user) loads CUDA.

I had a quick look at the code, and it seems this would require some refactoring since such extensions are not supposed to (and can't really) export new structs but are only intended to extend existing functionality in the main package. Without any more detailed inspection, I think it might be feasible to unify the XX and XXGPU structs by adding type parameters for the contained vectors, matrices, etc. I also wonder if it would be helpful to dispatch on the types of X and y instead of supporting a `device` keyword argument - it seems this would make it easier to extend methods such as `init_evotree` by only defining additional methods for `CuArray`s in the extension.

What do you think? Would you consider reviewing and possibly merging a PR with such changes?

devmotion:
I moved the CUDA-specific code to an extension. On Julia < 1.9, CUDA will continue to be a dependency and CUDA is supported automatically; on Julia >= 1.9, the package will not install CUDA by default but CUDA support can be enabled by loading CUDA (explicitly with e.g. `using CUDA` or when loaded by another dependency).

It was reasonably straightforward:
- I moved the CUDA-specific files to the ext/EvoTreesCUDAExt.jl directory
- Added name qualifications to (hopefully) everything in the extension that is not exported by EvoTrees
- Searched for "GPU", "Cu", and "CUDA" in the remaining src directory and moved GPU-specific branches to functions that can dispatch on the device type. It seems three functions (`device_ones` -> `ones` on the CPU, `CUDA.ones` on the GPU, `device_array_type` -> `Array` on the CPU, `CuArray` on the GPU, and `post_fit_gc` -> no-op on the CPU and garbage collection on the GPU) were sufficient.

The main problem is that I don't have access to an NVIDIA GPU, and hence can't verify that I did not break CUDA support 😅 I would recommend adding CUDA-specific tests with buildkite to the repo (see https://github.com/JuliaGPU/buildkite#adding-a-repository, the setup is quite straightforward, I did add it to a few repos successfully).

Fixes #226.


ext/EvoTreesCUDAExt/predict.jl
```diff
@@ -139,11 +127,11 @@ end
 
 function predict!(
```
jeremiedb, line 128: Would need `EvoTrees.predict!`

ext/EvoTreesCUDAExt/predict.jl
```diff
@@ -116,7 +104,7 @@ end
 # prediction from single tree - assign each observation to its final leaf
 function predict!(
```
jeremiedb, line 105: Would need `EvoTrees.predict!`

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -34,9 +34,9 @@ end
```
jeremiedb, line 34: Would need `EvoTrees.predict!` at line 30

Project.toml
```diff
@@ -37,4 +41,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 docs = ["Documenter"]
-test = ["DataFrames", "Test", "MLJBase", "MLJTestInterface"]
+test = ["CUDA", "DataFrames", "Test", "MLJBase", "MLJTestInterface"]
+
+[weakdeps]
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
```
jeremiedb, line 47: From Julia's formatter, it appears that expected order of project keys to be:
- deps
- compat
- weakdeps
- extensions
- compat
- extras
- targets

jeremiedb (changes requested):

jeremiedb:
Thanks for the PR!

1. I'm not clear whether CUDA should be kept in `deps`. It went fine on Julia 1.8, but I encountered an issue on 1.10 which I hadn't time to investigate
2. Regarding: https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/93, I'm not clear why the CI test did fail on nightly. My understanding is that CUDA should load fine regardless if the machine isn't NVIDIA compatible. Has there been a change to that in Julia 1.10?
3. Is the fit from EvoTrees poorer than for other classifiers in `rtstar` test? The need to pass `1_000` rather than the default 100 appears suspicious. I wouldn't expect a significant discrepancy compared to XGBoost (with its default to max_round=10 and learning_rate=0.3, I'd actually expects its fit to be inferior). If you'd have a data to share, I'd be curious to see if there might be an issue in the classifier routine.

devmotion:
> I'm not clear whether CUDA should be kept in deps. It went fine on Julia 1.8, but I encountered an issue on 1.10 which I hadn't time to investigate

Yes, CUDA should be kept in the `deps` section for backwards compatibility with Julia < 1.9. https://pkgdocs.julialang.org/v1/creating-packages/#Transition-from-normal-dependency-to-extension state explicitly

> Make sure that the package is both in the [deps] and [weakdeps] section. Newer Julia versions will ignore dependencies in [deps] that are also in [weakdeps].

Problems with 1.10 should be caused by something else.

Project.toml
devmotion: It depends on the Julia version you use: On Julia 1.6, the order Pkg uses is
- deps
- compat
- extensions
- extras
- targets
- weakdeps
whereas on Julia 1.7, 1.8, and 1.9 it is
- deps
- weakdeps
- extensions
- compat
- extras
- targets

See eg the discussion in https://github.com/JuliaTesting/Aqua.jl/issues/105 and in particular https://github.com/JuliaTesting/Aqua.jl/issues/105#issuecomment-1551221369

I guess you were referring to the order in Julia > 1.6 and accidentally included the compat section twice?

ext/EvoTreesCUDAExt/predict.jl
devmotion:
```suggestion
function EvoTrees.predict!(
```

ext/EvoTreesCUDAExt/predict.jl
devmotion:
```suggestion
function EvoTrees.predict!(
```

ext/EvoTreesCUDAExt/fit.jl
devmotion: Done.

devmotion:
> 2\. I'm not clear why the CI test did fail on nightly. My understanding is that CUDA should load fine regardless if the machine isn't NVIDIA compatible. Has there been a change to that in Julia 1.10?

No, it should work fine (the main problem with CUDA is that it pulls in many dependencies and in particular CUDA jlls). The problem with the current nightlies (https://github.com/TuringLang/MCMCDiagnosticTools.jl/actions/runs/6428712477/job/17456380934#step:6:335) is that CUDA defines GPU overrides for `Random.make_seed` (https://github.com/JuliaGPU/CUDA.jl/blob/da140359fc68f453e511608d9af552c90d699d9a/src/device/random.jl#L53) but `Random.make_seed` was removed last week (https://github.com/JuliaLang/julia/pull/51436).

Project.toml
jeremiedb: Yes I meant to refer to Julia > 1.6, sorry about the confusion for the duplicated `compat`.

jeremiedb:
Thanks for the clarifications. Regarding the Project.toml keys order, I'll likely move towards the Julia v1.6 order in subsequent release. Let me know if that's an issue for you. Or feel free to make an adjustment in this PR. 
Otherwise, good to go!

Project.toml
devmotion: I updated the order to the convention in Julia > 1.6.

devmotion:
I updated the order of the sections in the Project.toml file to the convention in Julia > 1.6.

jeremiedb (approved):

jeremiedb:
Registrator of new release v0.16.3 in underway. Thanks for your patience!

devmotion:
Thank you!

PR #262, opened 2023-10-17
Mlj table

### LINKED ISSUE #260: Categorical Variables Not Working
It looks like ordered categorical variables error:
```julia
tuned_machine = machine(tuning, hscf[:, [:Education, :Income]], hscf[:, :Age]) |> fit!
┌ Info: Training machine(DeterministicTunedModel(model = EvoTreeRegressor{EvoTrees.MSE}
│  - nrounds: 25
│  - lambda: 0.05
│  - gamma: 0.0
│  - eta: 0.04
│  - max_depth: 6
│  - min_weight: 1.0
│  - rowsample: 1.0
│  - colsample: 1.0
│  - nbins: 64
│  - alpha: 0.5
│  - monotone_constraints: Dict{Int64, Int64}()
│  - tree_type: binary
│  - rng: MersenneTwister(123)
└ , …), …).
[ Info: Attempting to evaluate 64 models.
┌ Error: Problem fitting the machine machine(EvoTreeRegressor{EvoTrees.MSE}
│  - nrounds: 25
│  - lambda: NaN
│  - gamma: 0.0
│  - eta: 0.04
│  - max_depth: 8
│  - min_weight: 1.0
│  - rowsample: 1.0
│  - colsample: 1.0
│  - nbins: 64
│  - alpha: 0.5
│  - monotone_constraints: Dict{Int64, Int64}()
│  - tree_type: binary
│  - rng: MersenneTwister(123)
│ , …). 
└ @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:682


1-element ExceptionStack:
AssertionError: T <: Real
Stacktrace:
  [1] get_edges(X::SubArray{Any, 2, Matrix{…}, Tuple{…}, false}; fnames::Vector{Symbol}, nbins::Int64, rng::MersenneTwister)
    @ EvoTrees ~/.julia/packages/EvoTrees/W4JTd/src/fit-utils.jl:8
  [2] init_core(params::EvoTreeRegressor{…}, ::Type{…}, data::SubArray{…}, fnames::Vector{…}, y_train::SubArray{…}, w::Vector{…}, offset::Nothing)
    @ EvoTrees ~/.julia/packages/EvoTrees/W4JTd/src/init.jl:4
  [3] init(params::EvoTreeRegressor{…}, x_train::SubArray{…}, y_train::SubArray{…}, device::Type{…}; fnames::Nothing, w_train::Nothing, offset_train::Nothing)
    @ EvoTrees ~/.julia/packages/EvoTrees/W4JTd/src/init.jl:222
  [4] fit(model::EvoTreeRegressor{…}, verbosity::Int64, A::@NamedTuple{…}, y::SubArray{…}, w::Nothing)
    @ EvoTrees ~/.julia/packages/EvoTrees/W4JTd/src/MLJ.jl:2
  [5] fit_only!(mach::Machine{…}; rows::Vector{…}, verbosity::Int64, force::Bool, composite::Nothing)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:680
  [6] fit_only!
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:606 [inlined]
  [7] #fit!#63
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:777 [inlined]
  [8] fit!
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:774 [inlined]
  [9] fit_and_extract_on_fold
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1235 [inlined]
 [10] (::MLJBase.var"#273#274"{MLJBase.var"#fit_and_extract_on_fold#300"{…}, Machine{…}, Int64})(k::Int64)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1062
 [11] _mapreduce(f::MLJBase.var"#273#274"{…}, op::typeof(vcat), ::IndexLinear, A::UnitRange{…})
    @ Base ./reduce.jl:440
 [12] _mapreduce_dim
    @ MLJBase ./reducedim.jl:365 [inlined]
 [13] mapreduce
    @ MLJBase ./reducedim.jl:357 [inlined]
 [14] _evaluate!(func::MLJBase.var"#fit_and_extract_on_fold#300"{…}, mach::Machine{…}, ::CPU1{…}, nfolds::Int64, verbosity::Int64)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1061
 [15] evaluate!(mach::Machine{…}, resampling::Vector{…}, weights::Nothing, class_weights::Nothing, rows::Nothing, verbosity::Int64, repeats::Int64, measures::Vector{…}, operations::Vector{…}, acceleration::CPU1{…}, force::Bool, per_observation_flag::Bool, logger::Nothing, user_resampling::CV)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1282
 [16] evaluate!(::Machine{…}, ::CV, ::Nothing, ::Nothing, ::Nothing, ::Int64, ::Int64, ::Vector{…}, ::Vector{…}, ::CPU1{…}, ::Bool, ::Bool, ::Nothing, ::CV)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1374
 [17] fit(::Resampler{CV, Nothing}, ::Int64, ::DataFrame, ::Vector{Int8})
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/resampling.jl:1535
 [18] fit_only!(mach::Machine{Resampler{…}, false}; rows::Nothing, verbosity::Int64, force::Bool, composite::Nothing)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:680
 [19] fit_only!
    @ MLJTuning ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:606 [inlined]
 [20] #fit!#63
    @ MLJTuning ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:777 [inlined]
 [21] fit!
    @ MLJTuning ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:774 [inlined]
 [22] event!(metamodel::EvoTreeRegressor{…}, resampling_machine::Machine{…}, verbosity::Int64, tuning::LatinHypercube, history::Nothing, state::@NamedTuple{…})
    @ MLJTuning ~/.julia/packages/MLJTuning/nZnsJ/src/tuned_models.jl:436
 [23] #35
    @ Base ~/.julia/packages/MLJTuning/nZnsJ/src/tuned_models.jl:474 [inlined]
 [24] iterate(g::Base.Generator, s::Vararg{Any})
    @ Base ./generator.jl:47 [inlined]
 [25] _collect(c::Vector{…}, itr::Base.Generator{…}, ::Base.EltypeUnknown, isz::Base.HasShape{…})
    @ Base ./array.jl:852
 [26] collect_similar
    @ MLJTuning ./array.jl:761 [inlined]
 [27] map
    @ MLJTuning ./abstractarray.jl:3275 [inlined]
 [28] assemble_events!(metamodels::Vector{…}, resampling_machine::Machine{…}, verbosity::Int64, tuning::LatinHypercube, history::Nothing, state::@NamedTuple{…}, acceleration::CPU1{…})
    @ MLJTuning ~/.julia/packages/MLJTuning/nZnsJ/src/tuned_models.jl:473
 [29] build!(history::Nothing, n::Int64, tuning::LatinHypercube, model::EvoTreeRegressor{…}, model_buffer::Channel{…}, state::@NamedTuple{…}, verbosity::Int64, acceleration::CPU1{…}, resampling_machine::Machine{…})
    @ MLJTuning ~/.julia/packages/MLJTuning/nZnsJ/src/tuned_models.jl:667
 [30] fit(::MLJTuning.DeterministicTunedModel{LatinHypercube, EvoTreeRegressor{…}}, ::Int64, ::DataFrame, ::Vector{Int8})
    @ MLJTuning ~/.julia/packages/MLJTuning/nZnsJ/src/tuned_models.jl:747
 [31] fit_only!(mach::Machine{…}; rows::Nothing, verbosity::Int64, force::Bool, composite::Nothing)
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:680
 [32] fit_only!
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:606 [inlined]
 [33] #fit!#63
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:777 [inlined]
 [34] fit!
    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:774 [inlined]
 [35] |>(x::Machine{MLJTuning.DeterministicTunedModel{LatinHypercube, EvoTreeRegressor{…}}, false}, f::typeof(fit!))
    @ Base ./operators.jl:917
```
Unordered categorical variables seem to error as well, but I'm not sure if that's because they're not supported or because of a bug.

jeremiedb:
Fix #260 
@ablaom I moved the MLJ wrapper from the orignal Matrix API to the new Tables one, which provides native support for richer inputs, incl. Ordered and Unordered (MultiClass) CategoricalValues.


ablaom:
Happy to review this shortly.

src/MLJ.jl
```diff
@@ -12,25 +17,6 @@ function okay_to_continue(model, fitresult, cache)
...
-MMI.reformat(::EvoTypes, X::AbstractMatrix, y) =
-  ((matrix=X, names=["feat_$i" for i = 1:size(X, 2)]), y)
-MMI.reformat(::EvoTypes, X::AbstractMatrix) =
-  ((matrix=X, names=["feat_$i" for i = 1:size(X, 2)]),)
-MMI.selectrows(::EvoTypes, I, A, y, w) =
-  ((matrix=view(A.matrix, I, :), names=A.names), view(y, I), view(w, I))
-MMI.selectrows(::EvoTypes, I, A, y) =
-  ((matrix=view(A.matrix, I, :), names=A.names), view(y, I))
-MMI.selectrows(::EvoTypes, I, A) = ((matrix=view(A.matrix, I, :), names=A.names),)
-
```
ablaom, line 33: My understanding is this: previously table -> matrix conversion happened in `reformat` and not in EvoTree's internal training, which means MLJ's `IteratedModel` wrapper avoided repeating these conversions over and over in controlled iteration. 

The change here suggests that table -> matrix conversion is now "internalised", which means the performance gain is lost, right? Or am I missing something.

jeremiedb: I'm afraid this brings back at the forefront the issue with regard to MLJ's handling of iterated process for models relying on caching / initialization mechanism. 

Once an EvoTree model is initialized with a training dataset, this dataset should be considered as fixed, no subsampling should be performed to it throughout the iterations. The algo is responsible for performing the subsampling operations throughout the iterations using tricks that are algo specific. This is true AFAIK for all "high performance" GBT models (incl. XGBoost, LightGBM, CatBoost). 

Looking at the error resulting when trying an IteratedModel uder this PR, why MLJ's `fit_only!` needs to perform a data resampling:  `fit(model, verbosity, _resampled_data(mach, model, rows)...)`?

The logic that has been used with EvoTrees so far to handle the MMI's `update` method has been to completely ignore the data args `A` and `y`: https://github.com/Evovest/EvoTrees.jl/blob/20b9e2a4a14863e87e16fe70fbece98c02b46ba2/src/MLJ.jl#L34
I'm still unclear why an issue happened with `IteratedModel` at first since `update` behavior where only the `model` and `cache` are actually used when iterating the model. And the EvoTrees' MLJ tests pass, so the basic operations seem to work fine, including iterative calls to `fit!`. 

This is still much WIP, but I'm trying to lay down what I have experiencced as shortcomings from ML toolkits, incl Scikitlearn, which assumes that a model can be instantiated with a `MyModelRegressor` rather than having distinct management of a model conguration, its initialization and the fitting of that instantiated model: https://github.com/Evovest/EvoTrees.jl/blob/learnAPI/experiments/learnAPI.jl. This is the kind of approach that we need to rely on internally to handle our ML pipelines, but I'd definitely like to see such support in an open framework.

jeremiedb: @ablaom The issue I encountered testing on `IteratedModel` appeared to be a red herring: the failure was caused by the `log_loss_` measure not accepting `Deterministic` output of the EvoTreeRegressor (isn't it unnecessarily restrictive? I actually don't see a logistic regression as less deterministic than "mse", but that's another topic!).  
It seems like the MLJ's `fit!` call to to reformat effectively results in a no-op as expected (each iteration following the 1st one directly mutate the instantiate fitted_model and the cache). Training speed is aligned with internal's `fit_evotree` if using very large evaluation `Step` (but quickly blows up, going from 12 sec reference time to ~68 sec with `Step(5)`). 

So back the this specific topic of moving the EvoTrees's table API, AFAICT the MLJ integration with MLJ works fine. Or at least as well as the existing one, but with added benefits in terms of richer input feature types.

ablaom:
Okay, I think I was mistaken to think dumping the model-specific data pre-processing (`reformat`) and model-specific resampling (`selectrows`) would slow down `IteratedModel`. As you point out, we are only calling `update` after the first train, so the conversion matrix -> table only happens once. I have also confirmed this with a benchmark, where this PR actually runs slightly faster for `IteratedModel`, for `Step` sizes  of 1, 5, 30 and a total of 300 iterations (details at the end). 

On the other hand, I do expect some impact on `TunedModel` wrapping, for here changing the parameter to be tuned means we call `fit` and not `update` for each new parameter value.  A benchmark suggest this is indeed the case, but the slowdown is  small (assuming it's even statistically significant):

```julia

# on EvoTrees master branch:

@btime fit!($mach, verbosity=0)
#  15.437 μs (100 allocations: 4.55 KiB)

# on this PR:

@btime fit!($mach, verbosity=0)
#  17.439 μs (100 allocations: 4.55 KiB)
```

Benchmark details are attached below.

So, I won't push back against your reluctance to update the `reformat/selectrows` implementation, and assume you  have good reasons for not doing so. 

I'm not sure I fully understand all the objections to the current API, but I suggest we have this discussion in a different forum. (Still happy to take a call to discuss this.) 

API changes to MLJModelInterface, if they are breaking, will not happen in any hurry. 

I will proceed with the rest of my review shortly. 

<details>

```julia
using MLJModels
using MLJIteration
using MLJBase
using BenchmarkTools
using StatisticalMeasures
using MLJTuning

EvoTreeBooster = @load EvoTreeRegressor
booster = EvoTreeBooster()

X, y = make_regression(1000, 5)


# # CONTROLLED ITERATION

# number of iterations:
N = 300

steps = [1, 5, 30]

iterated_boosters = [
    IteratedModel(
        model=booster,
        resampling=Holdout(fraction_train=0.8),
        controls=[Step(s), NumberLimit(round(Int, N/s))],
        measure=l1,
    )
    for s in steps
        ]

for ib in iterated_boosters
    mach = machine(ib, X, y)
    print("step=", ib.controls[1].n)
    @btime fit!($mach, verbosity=0)
    @assert report(mach).n_iterations == N
end

# on EvoTrees master branch:

# step=1  14.351 μs (101 allocations: 4.61 KiB)
# step=5  14.301 μs (101 allocations: 4.61 KiB)
# step=30  14.342 μs (101 allocations: 4.61 KiB)

# on this PR:

# step=1  13.701 μs (101 allocations: 4.61 KiB)
# step=5  13.726 μs (101 allocations: 4.61 KiB)
# step=30  13.776 μs (101 allocations: 4.61 KiB)


# # TUNING

r = range(booster, :lambda, lower = 0, upper = 0.5)
tuned_booster = TunedModel(
    booster,
    tuning=RandomSearch(),
    range=r,
    measure = l2,
    n = 500,
)

mach = machine(tuned_booster, X, y)

# on EvoTrees master branch:

@btime fit!($mach, verbosity=0)
#  15.437 μs (100 allocations: 4.55 KiB)

# on this PR:

@btime fit!($mach, verbosity=0)
#  17.439 μs (100 allocations: 4.55 KiB)
```
</details>

ablaom:
@jeremiedb There seems to be a problem with handling tables that are row-based. Perhaps you are calling `Tables.columnnames(X)` directly on a table `X` (a common gotcha)?? This only works for tables implementing column access.  I can look more carefully tomorrow. 

```julia
Pkg.activate("explore002", shared=true)

using MLJBase
using MLJModels
using Tables

EvoTreeBooster = @load EvoTreeRegressor
booster = EvoTreeBooster()

X, y = make_regression(1000, 5)

# this works:
mach = machine(booster, X, y) |> fit!

# this doesn't
X = Tables.rowtable(X)
mach = machine(booster, X, y) |> fit!

# ERROR: BoundsError: attempt to access 1000-element Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}} at index [1] 
# Stacktrace:
#  [1] getcolumn(x::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, i::Int64)                                                                                      
#    @ Tables ~/.julia/packages/Tables/NSGZI/src/Tables.jl:101
#  [2] fit(model::EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, verbosity::Int64, A::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, y::Vector{Float64}, w::Nothing)
#    @ EvoTrees ~/.julia/packages/EvoTrees/jzXEo/src/MLJ.jl:3
#  [3] fit(model::EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, verbosity::Int64, A::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, y::Vector{Float64})            
#    @ EvoTrees ~/.julia/packages/EvoTrees/jzXEo/src/MLJ.jl:3
#  [4] fit_only!(mach::Machine{EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, true}; rows::Nothing, 
# verbosity::Int64, force::Bool, composite::Nothing)                                         
#    @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:680
#  [5] fit_only!
#    @ ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:606 [inlined]
#  [6] #fit!#63
#    @ ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:777 [inlined]
#  [7] fit!
#    @ ~/.julia/packages/MLJBase/fEiP2/src/machines.jl:774 [inlined]
#  [8] |>(x::Machine{EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, true}, f::typeof(fit!))
#    @ Base ./operators.jl:907
#  [9] top-level scope
#    @ REPL[12]:1
```

jeremiedb:
> Perhaps you are calling `Tables.columnnames(X)` directly on a table `X`

Guilty :) The Tables implementation effectively assumed working with `columntables`. Since there were several dependencies to `getcolumn` internally, I've added a conversion to `Tables.columntable`. It doesn't add any overhead when Tables are already in such format which is the most common case I suppose (such as DataFrame). 

Regarding the impact on `TunedModel` I'd expect it to be very small, as when a new hyper-prams configuration is tested, a full model initialisation must be performed. Such initialization time is essentially the same for both the original Matrix and the new Tables based internal APIs. The bulk of computing time during an EvoTree training is normally spent iterating over n trees; steps which are identical in EvoTrees regardless in the input was originally a Matrix of a Tables. 

Thanks a lot for taking the time to look at the PR. And sorry about the digression about the iterated API. I'm trying to gather some coherent proposal so as to have a constructive take, will ping back once done!

src/fit-utils.jl
```diff
@@ -175,45 +175,40 @@ function split_set_threads!(
...
-                    x_bin,
-                    feat,
-                    cond_bin,
-                    feattype,
-                    offset,
-                    chunk_size,
-                )
-            end
-        end
+    @threads for bid = 1:nblocks
```
ablaom, line 178: This seems to go contrary to current advice: https://julialang.org/blog/2023/07/PSA-dont-use-threadid/ .

ablaom:
I think that row-table problem persists with `predict`:

```julia
EvoTreeBooster = @load EvoTreeRegressor
booster = EvoTreeBooster()

X, y = make_regression(1000, 5)
X = Tables.rowtable(X);

# smoke tests:
mach = machine(booster, X, y) |> fit!    # this now works, thanks
predict(mach, X) # fails

# ERROR: BoundsError: attempt to access 1000-element Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}} at index [1]    
```

<details>

<summary>Stack trace</summary>

```
Stacktrace:
 [1] getcolumn(x::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, i::Int64)                                                                                      
   @ Tables ~/.julia/packages/Tables/NSGZI/src/Tables.jl:101
 [2] binarize(df::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}; fnames::NTuple{5, Symbol}, edges::Vector{Any})                                                    
   @ EvoTrees ~/.julia/packages/EvoTrees/3Ocxy/src/fit-utils.jl:74
 [3] predict(m::EvoTrees.EvoTree{EvoTrees.MSE, 1}, data::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, ::Type{EvoTrees.CPU}; ntree_limit::Int64)
   @ EvoTrees ~/.julia/packages/EvoTrees/3Ocxy/src/predict.jl:91
 [4] predict(m::EvoTrees.EvoTree{EvoTrees.MSE, 1}, data::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}}, ::Type{EvoTrees.CPU})          
   @ EvoTrees ~/.julia/packages/EvoTrees/3Ocxy/src/predict.jl:83
 [5] predict(m::EvoTrees.EvoTree{EvoTrees.MSE, 1}, data::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}})                                
   @ EvoTrees ~/.julia/packages/EvoTrees/3Ocxy/src/predict.jl:83
 [6] predict(#unused#::EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, fitresult::EvoTrees.EvoTree{EvoTrees.MSE, 1}, A::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}})    
   @ EvoTrees ~/.julia/packages/EvoTrees/3Ocxy/src/MLJ.jl:45
 [7] predict(mach::Machine{EvoTrees.EvoTreeRegressor{EvoTrees.MSE}, true}, Xraw::Vector{NamedTuple{(:x1, :x2, :x3, :x4, :x5), NTuple{5, Float64}}})                                
   @ MLJBase ~/.julia/packages/MLJBase/fEiP2/src/operations.jl:133
 [8] top-level scope
   @ REPL[42]:1
```

</details>

jeremiedb:
> I think that row-table problem persists with `predict`:

Thanks for the catch. I had neglected adding a test for rowtables, which has now been added with within the MLJ test suite (both fit! and predict). Tests are now passing followin the patch to convert on `EvoTrees.predict`.

Refarding the `@spawn` vs `@thread`, I've followed the discussions around it. The post you referred to caused the migration to `:static` I did in v0.15, which turned out ill advised as it caused issues for a user running EvoTrees under an already threaded routine. My understanding of the issue is that problems can be caused when one relies on the `threadid()` as a reference to the thread making the work. This is to best of my knowledge not a pattern that has been used in the codebase. I simply wanted to move all threaded operation to a common appraoch, as the histogram builing routines we already using `@threads` and happenned to provide best performance.

src/MLJ.jl
```diff
@@ -47,40 +34,40 @@ function MMI.update(
     while cache[:info][:nrounds] < model.nrounds
       grow_evotree!(fitresult, cache, model)
     end
-    report = (features=A.names,)
+    report = (features=Tables.schema(A).names,)
```
ablaom, line 37: My understanding is that tables are not required to support `schema` directly. From the `Tables.Schema` docstring:

```
To get a table's schema, one can call Tables.schema on the result of Tables.rows or 
Tables.columns, but also note that a table may return nothing, indicating that its column
 names and/or column element types are unknown (usually not inferable). 
```

In the `fit` you are okay, because you first do `A -> Tables.columntable(A)`, but you didn't do that here. Maybe pass the names in `cache`, as they shouldn't change between `fit` and `update`.

src/MLJ.jl
```diff
@@ -47,40 +34,40 @@ function MMI.update(
...
   else
     fitresult, cache, report = fit(model, verbosity, A, y, w)
   end
   return fitresult, cache, report
 end
 
 function predict(::EvoTreeRegressor, fitresult, A)
-  pred = vec(predict(fitresult, A.matrix))
+  pred = predict(fitresult, A)
```
ablaom, line 45: This `pred` is a *vector*, and not a `1 x n` matrix, right? Otherwise, we need `vec` and I remember this being an issue in past.

ablaom: Similar queries apply at lines 50, 55, 60, 65, 70 below.

src/MLJ.jl
```diff
@@ -108,54 +95,39 @@ MMI.metadata_pkg.(
 
 MMI.metadata_model(
   EvoTreeRegressor,
-  input_scitype=Union{
-    MMI.Table(MMI.Continuous, MMI.Count, MMI.OrderedFactor),
-    AbstractMatrix{MMI.Continuous},
-  },
+  input_scitype=MMI.Table(MMI.Continuous, MMI.Count, MMI.OrderedFactor, MMI.Multiclass),
```
ablaom, line 98: Since you no longer support matrices I guess this is breaking, right?

And just to be sure, the treatment of `Multiclass` is properly invariant under relabelling? That is, the order of the classes makes no difference to outcomes? If not, then I suggest we rule out `MMI.Multiclass` and leave the responsibility of `Multiclass` -> `OrderedFactor` coercion to the user. 

I only ask, because a similar issue arose at DecisionTree.jl, where the ordering of categoricals matters,  but this was not flagged in docs.

src/fit-utils.jl
```diff
@@ -24,7 +24,7 @@ function get_edges(X::AbstractMatrix{T}; fnames, nbins, rng=Random.MersenneTwist
 end
 
 function get_edges(df; fnames, nbins, rng=Random.MersenneTwister())
-    _nobs = length(Tables.getcolumn(df, 1))
+    _nobs = Tables.DataAPI.nrow(df)
```
ablaom, line 27: Yeah, tables are not currently required to implement `DataAPI.nrow`, or something like it. This is a long standing issue. To be honest, I'd be happy to restrict to tables that that do implement `DataAPI.nrow` and push for implementations on a case-by-case basis. 

Of course my point is moot, if you only apply `get_edges` to column tables (output of `Tables.columntable`).

src/fit-utils.jl
ablaom: I've read your comment below and stand corrected. Thanks for the education 😄 .

ablaom (approved):
I have finished reviewing those parts of the PR that seemed relevant to the MLJ interface. It looks good to go.

src/MLJ.jl
jeremiedb: Yes EvoTrees' `predict` has been updated in the meantime to return vector rather than matrix for Regressor: https://github.com/Evovest/EvoTrees.jl/blob/535b0c47e7acb726e5d6591588deeceecadbacbe/src/predict.jl#L106

src/fit-utils.jl
jeremiedb: Thanks for pointing this out. Given that I enforce input columntable, then my understanding is this can be let as it is.

src/MLJ.jl
jeremiedb: The adoption of the Tables's API in EvoTrees was notably motivated by the support for actual `Unordered` factor. These receive a particular treatment in the establishing split points, so yes, they are order invariant. The unordered factors were already supported under the Matrix since they can be simply converted to their numeric levels and treated like normal ordered numerics. 

Since I'd rather not making this a breaking change, I've added a conversion from Matrix input to tables to `MMI.fit`, and adjusted the input scitypes accordingly, along a test with a matrix input.

src/MLJ.jl
jeremiedb: Done, using the `fnames` stored in `cache`.

PR #270, opened 2024-09-30
add FastTreeSHAPv1 implementation for EvoTrees

kpa28-git:
I was sad to find Julia doesn't have a TreeSHAP implementation so I ported [FastTreeSHAPv1](https://arxiv.org/abs/2109.09847) for my favorite boosted trees library. This PR is in case others find it useful.

This is marked as a draft for a few reasons:
1. To gauge interest to see if SHAP values should be in this package at all
2. The code runs but the results haven't yet been tested or compared to other SHAP value libraries yet
3. I'm not very familiar with the FastTreeSHAP algorithm, this was a simple 1:1 port of the pseudocode. So suggestions are very welcome
4. There may be a better way to parallelize this without using Channels. I usually use a `tmap` implementation for these kinds of parallel tasks, but I didn't want to add dependencies to the package just for this. Again, suggestions are welcome.

### DIFF
--- a/src/EvoTrees.jl
+++ b/src/EvoTrees.jl
@@ -45,6 +45,7 @@ end
 
 include("callback.jl")
 include("importance.jl")
+include("treeshapv1.jl")
 include("plot.jl")
 include("MLJ.jl")
 
--- a/src/treeshapv1.jl
+++ b/src/treeshapv1.jl
@@ -1,0 +1,144 @@
+"""
+FastTreeSHAPv1 SHAP values for a EvoTrees.EvoTree model
+
+## References
+* Jilei Yang, [*Fast TreeSHAP: Accelerating SHAP Value Computation for Trees*](https://arxiv.org/abs/2109.09847), 2022, pp. 16-17
+"""
+function treeshapv1(model::EvoTrees.EvoTree, x::AbstractVector; fnames=model.info[:fnames])
+    chan = Channel{typeof(x)}(length(model.trees))
+    @inbounds Threads.@threads for tree in model.trees
+        put!(chan, treeshapv1!(tree, zero(x), x))
+    end
+    ϕ = take!(chan)
+    while !isempty(chan)
+        ϕ += take!(chan)
+    end
+
+    pairs = string.(fnames) .=> ϕ
+    sort!(pairs, by=x->abs(x[2]), rev=true)
+    pairs
+end
+
+"""
+FastTreeSHAPv1 SHAP values for a single EvoTrees.Tree
+
+## References
+* Jilei Yang, [*Fast TreeSHAP: Accelerating SHAP Value Computation for Trees*](https://arxiv.org/abs/2109.09847), 2022, pp. 16-17
+"""
+function treeshapv1!(tree::EvoTrees.Tree, φ::AbstractVector, x::AbstractVector)
+    recurse!(tree, firstindex(tree.feat), PathValue[], Float64[], 1., PathValue(0, 1., true), φ, x)
+    φ
+end
+
+"""
+Element type used to make *m*, the path vector of unique features we have split on so far.
+
+## Fields
+* `d`: feature index of the jth internal node
+* `z`: covering ratio for the jth internal node
+* `o`: the threshold condition for the jth internal node
+"""
+struct PathValue
+    d::Int
+    z::Float64
+    o::Bool
+end
+
+function nodevalue(tree::EvoTrees.Tree{L, K}, j::Integer) where {L<:EvoTrees.GradientRegression, K}
+    tree.pred[1, j]
+end
+
+function nodevalue(tree::EvoTrees.Tree{L, K}, j::Integer) where {L<:Union{EvoTrees.LogLoss,EvoTrees.MLogLoss}, K}
+    sum(tree.pred[:, j]) # logit sum
+end
+
+function recurse!(tree::EvoTrees.Tree, j::Integer, m::AbstractVector{PathValue}, w::AbstractVector, q::Real, p::PathValue, φ::AbstractVector, x::AbstractVector)
+    m, w, q = extend(m, w, q, p)
+
+    if !tree.split[j] # leaf node
+        if length(m) > length(w)
+            s₀ = -sum(unwind(m, w, q)[2])
+        end
+
+        for i=2:lastindex(m)
+            vⱼ = nodevalue(tree, j)
+            if !m[i].o
+                update = s₀ * q * vⱼ
+            else
+                s = sum(unwind(m, w, q, i)[2])
+                update = s * q * (1 - m[i].z) * vⱼ
+            end
+            φ[m[i].d] += update
+        end
+    else
+        dⱼ = tree.feat[j]
+        aⱼ, bⱼ = (2j, 2j+1) # j's (left, right) child
+        h, c = x[dⱼ] ≤ tree.cond_float[j] ? (aⱼ, bⱼ) : (bⱼ, aⱼ) # (hot, cold) child
+        iz, io = 1, true
+        k = findfirst(x -> x.d == dⱼ, m)
+
+        if !isnothing(k)  # Undo previous updates if feature was already used
+            iz, io = m[k].z, m[k].o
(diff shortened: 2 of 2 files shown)


tecosaur:
You should be able to compare SHAP values to those from https://github.com/nredell/ShapML.jl, which has been tested against other implementations.

jeremiedb:
Sorry for late feedback, I think it would be a nice addition to the package to support direct SHAP derivation from the modeled tree. 

As @tecosaur mentionned, comparison with ShapML.jl would be a good preliminary step to asset the algo correctness. 

Then, regarding the interface of such SHAP calculation, it seems like in its current form, the implementation is intended to be applied on a single observation. I think it would be interesting to support an actual DataFrame/Table of many observations. That is to be coherent with the interface that accepts dataframe with features. that can be numeric, Bool, or Categorical. I shall be able to take a quick look at this in the upcoming days.

kpa28-git:
Cool. I don't have an immediate need for this anymore, but I'll do a comparison with an existing approximate SHAP value package sometime in the next couple weeks.

Your second point was my initial thought also. However, there doesn't seem to be a canonical way to aggregate SHAP values unless you know of one. A [ShapML.jl example](https://nredell.github.io/ShapML.jl/dev/) (see - Global feature importance heading) averages the absolute SHAP values for observations, which could be one generally useful way. I believe this would differ enough from the existing gain-based feature importance to be useful but I'd need to play with it to get an idea.

We can also plot all the shap values via a rug or beeswarm plot. I have some code to create rug plots to visualize missing data that can be adapted for shap values. This would be cool, but it depends on how much plotting you want in this package.

tecosaur:
Gentle poke @kpa28-git, you've clearly put a fair bit of work into this and it would be great to see it get over the line :slightly_smiling_face:

p.s. I have some SHAP-style beeswarm plotting code I'd be happy to share if it helps.

jeremiedb:
@kpa28-git I've attempted an adaptation of the SHAP calculations to the current state of the library.
I've encountered some issues, as some indexing on empty vector end up occuring. 

The updated implement lives under `src/shap.jl`: https://github.com/Evovest/EvoTrees.jl/blob/jdb/shap/src/shap.jl

You can replicate the issue encountered here:
https://github.com/Evovest/EvoTrees.jl/blob/fc826b5ef191911bd1daea8091744a5ff6bb44ac/experiments/shap.jl#L47-L50
I couldn't identify any obvious reason, like a some inconsistency. If you're interested validating the correctness of the `unwind` operation, that would much appreciated.

kpa28-git:
Yeah, I tried to do basic verification with iris a while ago but found the same bounds error. I don't think I got much (or any) further than you in understanding the bug, except that it's not necessarily limited to unwind. I'll need to go back to the paper.

Thanks for the work. I can't promise I'll be able to find the issue because of a big priority over the next two weeks, but I'll find the time to at least pull your changes and take a look.

jeremiedb:
Thanks for taking a look at it. I understand it comes with more overhead as it has been some time since the algo got implemented. 
I'd be happy to be able to get a TreeShap functionality embedded with EvoTrees. I wouldn't prioritize plotting funutionalities at the moment, as I think it would best fit with a dedicated XAI package.

kpa28-git:
Update: I was doing a search for other work on treeshap algorithms, and found another exact treeshap algorithm - [Linear TreeShap](https://arxiv.org/abs/2209.08192). I don't know how I missed it last year, but it has better time complexity than fasttreeshapv1/v2 and is more straightforward algorithmically.

I've started working on an implementation. There are still some details to work out until I get to something testable, but I think this is a better path.

kpa28-git:
Update: had some time to work on this today. It's mostly implemented (essentially a port of the [paper author's implementation](https://github.com/yupbank/linear_tree_shap/tree/main/linear_tree_shap)), except for one thing.

I need the number of training samples that reach the ith node (something like [`n_node_samples[i]` in sklearn's Tree](https://scikit-learn.org/stable/auto_examples/tree/plot_unveil_tree_structure.html#sphx-glr-auto-examples-tree-plot-unveil-tree-structure-py)). This doesn't exist in the `Tree` or info dict, I haven't looked for this much in the rest of the library yet. For now, I'll probably run through the training data again after `fit` to get this.

Other than that, I tested the polynomial math functions against the reference a bit and the outputs were the same. Actual algorithm and weight / edge_height functions can't be tested until I have the `n_node_samples` thing.

jeremiedb:
Thanks for the update! 

Regarding the need for the number of samples reaching any given node, could you confirm whether it's the just `n_node_samples` and not `weighted_n_node_samples`? If it's the later (weighted observations), it would actually make things easier as the the sum of weights is an information the would be readily available to feed to the tree during training. 

The approach I'd take is to add a field to the `Tree` struct: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/structs.jl#L57-L63, which would be populated at the creation of the tree:
https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit.jl#L93-L98

I could take a shot in the next few days to have such (weighted) node observations added in Tree structure on a distinct branch. Let me know if that would be useful to you.

kpa28-git:
Thanks, I'm going based on [the author's implementation here](https://github.com/yupbank/linear_tree_shap/blob/main/linear_tree_shap/utils.py#L16)  (uses `tree.n_node_samples`). From a quick look, it doesn't look like the author uses any sample weighting in their [test runs](https://github.com/yupbank/linear_tree_shap/tree/main/tests) though. I don't know if you've seen any different?

So for their purposes, it would be the same to use the `n_node_samples` or `weighted_n_node_samples`.

Also, I don't think this changes anything but all I need is the ratio of `tree.n_node_samples[node] / tree.n_node_samples[parent]`, not the actual values themselves.

Just for testing, I was thinking to get the node weights after fitting (to be able to test shap values for at least the unweighted case). But, since the node sample ratios of unweighted and "weighted" would be the same for the unweighted case that would also work.

The only question is if I need to use weighted node sample ratio for the weighted case. That intuitively sounds correct, but I'd need to go back to the paper or test both ways to be sure. Assuming that's true, your idea would work. I have to look into it more. Right now, I'll focus on getting the unweighted case working.

yupbank:
Yes, weighted_n_node_samples would do. Unweighted essentially have an implicit weight of 1. And what matters is the ratio of weights.

jeremiedb:
@yupbank Thanks for your input!

@kpa28-git Adding weights to the Tree struct was really triviail. Following branch `shap-w` includes the information in the trained model: https://github.com/Evovest/EvoTrees.jl/tree/jdb/shap-w.

```julia
config = EvoTreeRegressor(; max_depth=3)
model = fit(config, dtrain; target_name="y")

julia> model.trees[2].w
7-element Vector{Float32}:
 8000.0
 5250.0
 2750.0
 4000.0
 1250.0
 1125.0
```

Note that first Tree of the model should be ignored by the Shap procedure as it's only used to set the bias, so `w` is 0.0:
```julia
julia> model.trees[1]
EvoTrees.Tree{EvoTrees.MSE, 1}
 - feat: [0]
 - cond_bin: UInt8[0x00]
 - w: Float32[0.0]
 - gain: Float32[0.0]
 - pred: Float32[0.56130254;;]
 - split: Bool[0]
```

Let mw know if that fixes your need for node weight information.

kpa28-git:
@yupbank Appreciate it! Nice paper by the way

@kpa28-git Thanks! Will check it out and get back to you after I can start testing the algo.

kpa28-git:
Funny to see you're both in Montréal. Small world.

jeremiedb:
@kpa28-git For info, note that the weight info in the resulting fitted trees (`tree.w`) is now on main and included in latest release (v0.18.0). Branch `jdb/shap-w` is thus no longer needed and has been deleted.

kpa28-git:
@jeremiedb Great, my port of linear TreeShap wasn't consistent with ShapML for deeper trees and I hadn't gotten around to fixing it.

I've no particular interest in FastTreeShap, closing the draft.

PR #294, opened 2025-08-11
Improving GPU Performance

### LINKED ISSUE #288: Improve GPU performance
This issue is in relation with SciML small grants program. 

The objective of this project is to improve the GPU backend to bring the training benchmarks to in the competitive range to XGBoost. 

We want to bring training time on GPU to not more than 25% that of XGBoost on the benchmarks: https://github.com/Evovest/EvoTrees.jl/blob/main/benchmarks/regressor.jl

### Current benchmarks

device | nobs | nfeats | max_depth | train_evo | train_xgb | diff
-- | -- | -- | -- | -- | -- | --
gpu | 1000000 | 10 | 6 | 2.2 | 1.0 | 111%
gpu | 1000000 | 10 | 11 | 24.3 | 3.1 | 679%
gpu | 1000000 | 100 | 6 | 3.6 | 3.2 | 13%
gpu | 1000000 | 100 | 11 | 31.1 | 8.5 | 265%
gpu | 10000000 | 10 | 6 | 8.4 | 7.8 | 8%
gpu | 10000000 | 10 | 11 | 42.1 | 13.6 | 209%
gpu | 10000000 | 100 | 6 | 21.7 | 28.9 | -25%
gpu | 10000000 | 100 | 11 | 68.6 | 57.8 | 19%

Note that currently, the desired performance would already be met for some scenarios. E.g.: 10M observations, 100 feature, and depth of 6 or 11.

However, the performance gap is important for larger depths and smaller number of observations. E.g.: 6.79 times slower for depth 11, 1M observations and 10 features.

A key bottleneck is expected to originate from the important overhead associated at the per node level at larger depths, as their number grows as 2^depth. Therefore, a large number of kernels is launched as the depth of the tree grows. Also, only the gradients
and histograms are currently computed on the GPU, while gains and best node split could also be computed on the GPU
and limit the GPU to CPU communications.

### Existing work
Path for improvements has been developed in https://github.com/Evovest/EvoTrees.jl/tree/gpu-hist

device | nobs | nfeats | max_depth | train_evo | train_xgb | diff
-- | -- | -- | -- | -- | -- | --
gpu | 1000000 | 10 | 6 | 1.6 | 1.0 | 68%
gpu | 1000000 | 10 | 11 | 4.8 | 2.7 | 75%
gpu | 1000000 | 100 | 6 | 3.9 | 2.9 | 33%
gpu | 1000000 | 100 | 11 | 11.4 | 8.0 | 43%
gpu | 10000000 | 10 | 6 | 11.0 | 7.6 | 45%
gpu | 10000000 | 10 | 11 | 21.6 | 13.2 | 64%
gpu | 10000000 | 100 | 6 | 39.1 | 27.3 | 43%
gpu | 10000000 | 100 | 11 | 81.2 | 51.0 | 59%

Its approach is to build histograms and gains as a single Arrays for the full tree, allowing a single kernel call at each depth, regardless of the number of nodes. 

The other important design change is that instead of tracking a vector of observations ids for each node, a single vector of length == length(obs) is used to track the node id to which each observation is now assigned: https://github.com/Evovest/EvoTrees.jl/blob/c8694f7237f98f0cbf004866f8d92590169d47fb/ext/EvoTreesCUDAExt/fit-utils.jl#L127

However, this work is not leveraging the histogram subtraction trick: the histogram with largest number of observations can be derived by subtracting the other child histogram to the parent one. It's not clear if the subtraction trick would still be useful with this design where each observation needs to be iterated to know to which node it belongs.

### General considerations
There's no hard requirement about keeping the current design where each node is treated as a distinct entity or adopting the `gpu-hist` WIP work where a single kernel can be launched per depth. 

Both the improvements to overhead costs associated with the large number of nodes and the change in structure to have the computation cost to grow linearly with depth may be sound design; and the mixing both ideas may be possible.

Having the GPU computation performed through [KernelAbstractions.jl])https://github.com/JuliaGPU/KernelAbstractions.jl) would be a nice to have, notably to allow AMD GPU support.

AdityaPandeyCN:
This PR adds initial changes to resolve #288 by implementing comprehensive GPU optimizations targeting to bring EvoTrees.jl's training performance within 25% of XGBoost on GPU, while adding AMD GPU support( and other types) through KernelAbstractions.jl.
This is an intial draft to be updated with local benchmarking


AdityaPandeyCN:
Hello @jeremiedb. Can you have a look?

jeremiedb:
Thanks for the PR!
I'm out in holidays right now, I'll be able to take a closer look by end of the weekend.

AdityaPandeyCN:
No problem

jeremiedb:
Hi @AdityaPandeyCN, some comments:

1. For proper Proejct.toml dependency setup, could you consider the changes done in https://github.com/Evovest/EvoTrees.jl/pull/295/commits/c14df49f21a88463d99a2dfbf2b496006c80e8c1#diff-72ed386c2a0cd1d23c0968297e70023ed98c22490d146dd89fc91f48369bad4d
2. The above mai be related to the more general consideration that this PR is open against https://github.com/Evovest/EvoTrees.jl/tree/gpu-hist rather than `main`. There have been several changes/fixes between `gpu-hist` and `main`, and ultimately, the GPU should be merged on main. Note that in `gpu-hist`, both the CPU and GPU structures follow a different histogram structure than in current main: a single large histogram object than a vector of per node histogram. And under `gpu-hist`, this single 4D histogram resulted in substantial slowdown compared to main. The GPU improvement should not come at the cost of a regression on CPU performance. This could be achieved either by:
   
   a) keep the CPU structure currently in `main`, resulting in more structural differences between `cpu` and `gpu` implementations. This is not my preferred scenario, but would be acceptable. 
   
   b) perform adaption on CPU side to maintain performance compared to current `main`. My understanding is that the performance regression on CPU on `gpu-hist` side is mostly cause by the core histogram accumulation. These were likely due to lack of histogram subtration (which I think you resolved), as well as sub-performant code generation which I believe didn't end-up with simd unlike in handcrafted cpu histogram accumulatin in `main`: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit-utils.jl#L303

3. Related to the above alignment with the state of the features in latest release (ie. `main`), make sure that the histogram notably support the `monotonic_constraints`: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit-utils.jl#L419-L425
as well as the `unordered categorical variables: 
https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit-utils.jl#L476-L481

4. Trying to run with a CUDA benchmark results in an error: https://github.com/Evovest/EvoTrees.jl/blob/c14df49f21a88463d99a2dfbf2b496006c80e8c1/benchmarks/regressor.jl
```julia
ERROR: MethodError: no method matching zeros(::CUDABackend, ::Int64, ::Int64)
```
This seems related to missing `Type` as a 2nd arg in https://github.com/Evovest/EvoTrees.jl/blob/c14df49f21a88463d99a2dfbf2b496006c80e8c1/ext/EvoTreesCUDAExt/fit.jl#L48-L49
As I understand that you have access to a local GPU, could you ensure that `regressor.jl` benchmark runs successfully when pushing a PR?

Let me know if the above comments make sense to you. If you feel that are many ambiguities, we can make a quick clarification call. Although there I see there are still quite a few integraton work remaining, I think it would make sense to first have the benchmark able to run so as to be able to assess whether the work you've done actually manage to hit the performance target on the simple `regressor.jl` benchmark that only contains numeric feature without monotonic constraints. If performance is conclusive, I'm confident the adaptations won't be a significant blocker as these are non compute intensive ops.

AdityaPandeyCN:
Hello @jeremiedb,  I just wanted to inform you that I will take 1-2 day more to come up with the updated code for review. I am able to successfully run the benchmarks, but the performance time are not what we want. I have identified the issues and in a correct path to solve it but would want some more time to implement everything accurately. I am sorry to keep you in a hold like this, I will update you with the results.

jeremiedb:
Thanks for the follow up. No worry about the timing, it's totally understandable as it's not a trivial PR to adress. 
I'm actually positive about the progress made so far and that you've tackled it with backend-agnostic KA.jl.

AdityaPandeyCN:
Hello @jeremiedb, I have fixed the benchmarks and they are performing slightly better than the main branch code. Although this still has not hit the target( on my machine), I thought it is better to have you take a look. I will do some more improvements which I am confident of hitting our targets, but it will be better if you watch the progress. My regressor.jl is for a specific case.

AdityaPandeyCN:
Thank you for this, I apologise for completely ignoring CPU, I will fix this. Regarding the performance I realised the atomic contention in `hist _kernel` was providing some bottleneck, I optimised it and got some more improvements ( I have not pushed it yet in the PR). 
Following you advice I will run the benchmark ensuring both cpu and gpu performance and will update you.

jeremiedb:
Excellent; good luck on the next steps!

AdityaPandeyCN:
What do you think I should do? I am thinking of the solution where I modify the main branch code to make them faster and get the target 25% benchmark goal but without using KA.jl like pure CUDA as it was, this code with KA.jl stays in the gpu-hist targetting the 25% goal but slow cpu  performance( I will keep working on it to make the cpu  performance  better).

jeremiedb:
I've run complete benchmarks on a RTX A4000 and the results look very good, well done! https://github.com/Evovest/EvoTrees.jl/pull/295/files#diff-366df6eb18792d86bc2727bddcb127e48186f82968e727ebd703311331bfd59c
The only sibstantially slower scenario was on 10K, 100 feat, depth 11; which was anyway outside of the state scope, and which would be more appropriate to run on CPU.

On the cpu side, the slowdown is effectively substantial compared to `main` https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu/benchmarks/results/regressor-cpu.csv
It's quite aligned with what I had initially observed with these 4D histograms: https://github.com/Evovest/EvoTrees.jl/tree/gpu-hist?tab=readme-ov-file#cpu

As far as GPU speed is conerned, the objective would be met. I think that integrating this new GPU design along the current CPU on `main` is a good direction. It could be eventually revisited if adopting the same structure on CPU as for GPU can work out. 

There are however a couple of features that would need to be integrated; which I previously alluded to. 
Without being exhaustive list, I can think of:
- support for monotonic contraint: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit-utils.jl#L419-L425
- Handling of unorderd categorical variable: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit-utils.jl#L476-L481
- usage of dedicated struct to store cache: https://github.com/Evovest/EvoTrees.jl/blob/main/ext/EvoTreesCUDAExt/structs.jl
- Need to support the multi-classification loss and gaussiaMLE, which aren't limited to 3 elements (2 grads + weights). This PR hard-coded the 3 elements: 
- https://github.com/AdityaPandeyCN/EvoTrees.jl/blob/9f687a414706ff084fbe9726e46bc3bb1bb8839c/ext/EvoTreesCUDAExt/fit-utils.jl#L72-L74
- https://github.com/AdityaPandeyCN/EvoTrees.jl/blob/9f687a414706ff084fbe9726e46bc3bb1bb8839c/ext/EvoTreesCUDAExt/fit-utils.jl#L107-L109
- https://github.com/AdityaPandeyCN/EvoTrees.jl/blob/9f687a414706ff084fbe9726e46bc3bb1bb8839c/ext/EvoTreesCUDAExt/fit-utils.jl#L124-L126
You can see that this is currently handled in `main` branch with a thread dimension dedicated to handling these K possible values to be added: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/ext/EvoTreesCUDAExt/fit-utils.jl#L17 Perhaps a loop should be used over the K items (instead of current hard-coded 3)?
I see that the proper dimensions of the cached assets should already be in place with regard to K, as seen in `init.jl`:
https://github.com/AdityaPandeyCN/EvoTrees.jl/blob/9f687a414706ff084fbe9726e46bc3bb1bb8839c/ext/EvoTreesCUDAExt/init.jl#L66
I think that these initialisations should also be converted from `CUDA.xxx` to `KA.xxx`?

Let me know if you're aligned you the remaining steps.

AdityaPandeyCN:
> * With KA.jl, when specifying the `ndrange`, there's no need to worry about busting the max CUDA gridsize of 65,535?

I think KA.jl handles this through its NDIteration.partition function(https://github.com/JuliaGPU/KernelAbstractions.jl/blob/b525d6bbdc15d76c74b4d7ffbd4314c88f3adb93/src/KernelAbstractions.jl#LL719-L721)

AdityaPandeyCN:
Hello @jeremiedb I have been sick for few days so couldn't get significant progress.I was doing the integration of my code with the main branch but this is not going as planned, I feel there are many things that need to be properly integrated. Do you think I should work on adding the missing features you listed here? I can maybe than just improve the cpu performance on my branch code itself.

AdityaPandeyCN:
Hello @jeremiedb, I started with replacing the whole ext files with our gpu-hist branch one as you have also written but I think  

>then let the calls to `init` and `fit` from core `fit` functions (
> 
> https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit.jl#L321
> 
> ,
> https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit.jl#L335
> 
> ) handle the dispatch to your updated GPU support. Then, the remaining steo would be to integrate the missing features previously discussed.

This is where I did mistake and was not able to integrate well, but again I was not doing well so maybe I was not doing this with full focus. I think I can do it now, you are right that improving cpu performance on this branch would be unnecessarily complex.

>Let me know what you think. I may be able to take some time this weekend to take a quick shot at integrating you GPU extension into main, and better understand what kind of unexpected issues there may be.

I will work on this again, I have made some significant progress on the features additions and think that the integration would be achievable now. I will update you with the progress.

AdityaPandeyCN:
Closing this with a follow up PR to the main branch with integrated code.

PR #299, opened 2025-08-31
Improving training performance on GPU backend

AdityaPandeyCN:
This integrates changes introduced in this PR(https://github.com/Evovest/EvoTrees.jl/pull/294) to improve training performance on GPU backend  which targets to solve this issue(https://github.com/Evovest/EvoTrees.jl/issues/288).

### DIFF
--- a/ext/EvoTreesCUDAExt/fit.jl
+++ b/ext/EvoTreesCUDAExt/fit.jl
@@ -1,22 +1,22 @@
-function EvoTrees.grow_evotree!(evotree::EvoTree{L, K}, cache::CacheGPU, params::EvoTrees.EvoTypes) where {L, K}
-	EvoTrees.update_grads!(cache.∇, cache.pred, cache.y, L, params)
+function EvoTrees.grow_evotree!(evotree::EvoTree{L,K}, cache::CacheGPU, params::EvoTrees.EvoTypes) where {L,K}
+    EvoTrees.update_grads!(cache.∇, cache.pred, cache.y, L, params)
 
-	for _ in 1:params.bagging_size
-		is = EvoTrees.subsample(cache.is_in, cache.is_out, cache.mask, params.rowsample, params.rng)
+    for _ in 1:params.bagging_size
+        is = EvoTrees.subsample(cache.is_in, cache.is_out, cache.mask, params.rowsample, params.rng)
 
-		# Feature sampling done on CPU then copied to GPU
-		js_cpu = Vector{eltype(cache.js)}(undef, length(cache.js))
-		EvoTrees.sample!(params.rng, cache.js_, js_cpu, replace = false, ordered = true)
-		copyto!(cache.js, js_cpu)
+        # Feature sampling done on CPU then copied to GPU
+        js_cpu = Vector{eltype(cache.js)}(undef, length(cache.js))
+        EvoTrees.sample!(params.rng, cache.js_, js_cpu, replace=false, ordered=true)
+        copyto!(cache.js, js_cpu)
 
-		tree = EvoTrees.Tree{L, K}(params.max_depth)
-		grow_tree!(tree, params, cache, is)
-		push!(evotree.trees, tree)
-		EvoTrees.predict!(cache.pred, tree, cache.x_bin, cache.feattypes_gpu)
-	end
+        tree = EvoTrees.Tree{L,K}(params.max_depth)
+        grow_tree!(tree, params, cache, is)
+        push!(evotree.trees, tree)
+        EvoTrees.predict!(cache.pred, tree, cache.x_bin, cache.feattypes_gpu)
+    end
 
-	evotree.info[:nrounds] += 1
-	return nothing
+    evotree.info[:nrounds] += 1
+    return nothing
 end
 
 """
@@ -25,13 +25,13 @@ end
 Grow an oblivious tree on GPU (currently falls back to standard binary tree with a warning).
 """
 function grow_otree!(
-	tree::EvoTrees.Tree{L, K},
-	params::EvoTrees.EvoTypes,
-	cache::CacheGPU,
-	is::CuVector,
-) where {L, K}
-	@warn "Oblivious tree GPU implementation not yet available, using standard tree" maxlog = 1
-	grow_tree!(tree, params, cache, is)
+    tree::EvoTrees.Tree{L,K},
+    params::EvoTrees.EvoTypes,
+    cache::CacheGPU,
+    is::CuVector,
+) where {L,K}
+    @warn "Oblivious tree GPU implementation not yet available, using standard tree" maxlog = 1
+    grow_tree!(tree, params, cache, is)
 end
 
 """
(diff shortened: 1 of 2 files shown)


AdityaPandeyCN:
Hello @jeremiedb I have done the integration, but I think this will need some more iteration from me as I may have missed some bits.I can see slight regression in our GPU Performance after this integration but it can be improved, moreover both the gpu and cpu benchmarks are working correctly as of now.

jeremiedb:
Thanks! I'll get back to you within a day, so I can run benchmarks and review potential missing features.

jeremiedb:
Performance wise, this branch meets the target, and I actually observed slight improvement vs the previous PR: 
https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu-v2/benchmarks/results/regressor-gpu.csv

To get a high level assessment of correctness, I ran the some basic fitting used to generate docs support: 
https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu-v2/experiments/readme_plots.jl
Some issues were encountered for losses that otherwise work properly for GPU on `main` branch: 
- `mae` loss fails: https://github.com/Evovest/EvoTrees.jl/blob/40fcb3074f11296e5846a30dfead996a0efc3198/experiments/readme_plots.jl#L185-L194
- quantile loss fails: https://github.com/Evovest/EvoTrees.jl/blob/40fcb3074f11296e5846a30dfead996a0efc3198/experiments/readme_plots.jl#L337-L364
- credibilty-based losses fails https://github.com/Evovest/EvoTrees.jl/blob/40fcb3074f11296e5846a30dfead996a0efc3198/experiments/readme_plots.jl#L443-L497
- Gaussian trained without error, but the resulting fitted model shows abnormal pattern. From this PR: https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu-v2/docs/src/assets/gaussian-sinus-binary-gpu.png vs from `main`: https://github.com/Evovest/EvoTrees.jl/blob/main/docs/src/assets/gaussian-sinus-binary-gpu.png
I didn't have a proper benchmark for multi-classification/softmax, my initial hypothesis would be that some there may have been incorrectness in the handling of the accumulation where there are more than a single parameters for which the gradiants/hessian need to be computed.

jeremiedb (commented):
Some comments concern the extent to which your aiming a full support generic backend through KernelAbstractions, as there are several operationsthat remain specific to CUDA. As mentionned in the comment, I'm okay to move with an incomplete transition to KA, while keeping CUDA specifics to start with. If you think it's reasonable to move the remaining CUDA elements to KA, that would be ideal though.  

Other key comments concern missing support for some losses. These would need to be adressed. I haven't tested yet the support for monotonic constraint and categorical features. If all of these work properly, then we should be pretty much done!

AdityaPandeyCN:
Thankyou @jeremiedb for running the benchmarks, some of the omitted supports were primarily due to me replacing them with the gpu-hist files(loss.jl and predict.jl). I know how to workaround it know and will update you  with the complete implementation in some days.

jeremiedb:
@AdityaPandeyCN I've put most comments in https://github.com/Evovest/EvoTrees.jl/pull/300, which I thought may help highlight the issues I'm seeing with performance regression that got introduced.

AdityaPandeyCN:
@jeremiedb Hello, Can you have a look at the code I have tried fixing them and can see the performance back. I think I will be happy with just a code review from your side and we can run the benchmarks again once I address your points.

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
-                    offset += length(nodes[n].is)
+    copyto!(tree.split, Array(cache.tree_split_gpu))
+    copyto!(tree.feat, Array(cache.tree_feat_gpu))
+    copyto!(tree.cond_bin, Array(cache.tree_cond_bin_gpu))
+    copyto!(tree.gain, Array(cache.tree_gain_gpu))
 
-                    nodes[n<<1].is, nodes[n<<1+1].is = _left, _right
-                    nodes[n<<1].∑ .= nodes[n].hL[:, best_bin, best_feat]
-                    nodes[n<<1+1].∑ .= nodes[n].hR[:, best_bin, best_feat]
+    leaf_nodes = findall(x -> !tree.split[x] && x > 0, 1:length(tree.split))
```
jeremiedb, line 158: isn't `leaf_nodes = findall(x -> !x, findall)`  all that's needed?

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
-    x_bin::CuMatrix,
-    feattypes::Vector{Bool},
-    monotone_constraints,
-) where {L,K,N}
-
-    # initialize
-    n_current = [1]
-    depth = 1
-    jsg = CuVector(js)
+    epsv = eltype(tree_pred)(1e-8)
```
jeremiedb, line 214: why use `epsv` rather than `eps`?

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
-            sort!(n_current)
-            @threads for n ∈ n_current
-                update_gains!(L, nodes[n], js, params, feattypes, monotone_constraints)
+            gain_l = zero(eltype(nodes_gain))
+            gain_r = zero(eltype(nodes_gain))
+            
+            @inbounds for k in 1:K_val
+                g_l = nodes_sum[k, child_l]
+                h_l = nodes_sum[K_val+k, child_l]
+                gain_l += g_l^2 / (h_l + lambda * w_l / K_val + epsv)
```
jeremiedb, line 255: Why the divisions by `K_val`? There's no such operation in the current CPU or GPU implementations. Is there redundant addition of the weight that gets performed earlier?

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
             end
-            for n ∈ n_current
-                if n > 1
-                    @views nodes[1].gains[:, js] .*= nodes[n].gains[:, js] .> 0 #mask ignore gains if any node isn't eligible (too small per leaf weight)
+        else
+            w = nodes_sum[2*K_val+1, node]
+            @inbounds for k in 1:K_val
+                g = nodes_sum[k, node]
+                h = nodes_sum[K_val+k, node]
+                if w <= zero(w) || h + lambda * w / K_val <= zero(h)
```
jeremiedb, line 287: same as above, the ` / K_val` seems a suspicious operation

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
+            w = nodes_sum[2*K_val+1, node]
+            nodes_gain[node] = g^2 / (h + lambda * w + eps)
+        else
+            gain_sum = zero(T)
+            w = nodes_sum[2*K_val+1, node]
+            
+            @inbounds for k in 1:K_val
+                g = nodes_sum[k, node]
+                h = nodes_sum[K_val+k, node]
+                gain_sum += g^2 / (h + lambda * w / K_val + eps)
```
jeremiedb, line 322: Same `/K_val` comment. 

Also, why not use `max(eps, ...)` like in `cpu` implementation?

ext/EvoTreesCUDAExt/init.jl
```diff
@@ -25,7 +22,6 @@ function EvoTrees.init_core(params::EvoTrees.EvoTypes, ::Type{<:EvoTrees.GPU}, d
     elseif L == EvoTrees.MLogLoss
         if eltype(y_train) <: EvoTrees.CategoricalValue
             target_levels = EvoTrees.CategoricalArrays.levels(y_train)
-            target_isordered = EvoTrees.isordered(y_train)
```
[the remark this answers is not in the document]
jeremiedb: `target_isordered` should be kept

jeremiedb:
Hi @AdityaPandeyCN, I've left some comments. There seems to be some correctness issues remaining, notably for non-standard losses like `gaussianMLE` or `quantile`. 
This can notably be seen by plotting the fitted preidictions here: https://github.com/Evovest/EvoTrees.jl/blob/ead79c5824d8593d96ce0fb00bd7a1b74f944a28/experiments/readme_plots.jl#L254
https://github.com/Evovest/EvoTrees.jl/blob/ead79c5824d8593d96ce0fb00bd7a1b74f944a28/experiments/readme_plots.jl#L335

The chart exhibits some broken pattern that contrast with the equivalent figures run on `cpu` (changin the device to `:cpu` https://github.com/Evovest/EvoTrees.jl/blob/ead79c5824d8593d96ce0fb00bd7a1b74f944a28/experiments/readme_plots.jl#L15). These issues are likely tied the above mentionned conerns in some of the computations.

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
+function grow_otree!(
+    tree::EvoTrees.Tree{L,K},
+    params::EvoTrees.EvoTypes,
+    cache::CacheGPU,
+    is::CuVector
+) where {L,K}
+    @warn "Oblivious tree GPU implementation not yet available, using standard tree" maxlog=1
+    grow_tree!(tree, params, cache, is)
+end
+
```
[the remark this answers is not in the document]
AdityaPandeyCN: True GPU oblivious remains unimplemented due to kernel complexities.

AdityaPandeyCN:
Hello @jeremiedb I have tried to address all the comments. My  target is to perform the computations on the gpu as much as we could avoiding cpu fallback.

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,361 @@
...
-        n_next = Int[]
-
-        if depth == params.max_depth
-            for n in n_current
-                if L <: Quantile
-                    pred_leaf_cpu!(tree.pred, n, nodes[n].∑, L, params, ∇, nodes[n].is)
-                else
-                    pred_leaf_cpu!(tree.pred, n, nodes[n].∑, L, params)
+        # Compute child gradient sums from histogram (enables subtraction at next level!)
+        # This is CRITICAL: stores gradient sums so next level can use:
```
jeremiedb, line 301: Why mentions that `child gradient sums`  is necessary for `subtract_hist_kernel!`? The later only depends on `h∇` to perform the subtraction, per: https://github.com/Evovest/EvoTrees.jl/pull/299/files#diff-d696002f188318b01ef653ce29625f5597f4bc2f629f9885226494cc4286fd18R121-R124

jeremiedb:
@AdityaPandeyCN Thanks! I've noticed some regression on performance as showned in: 
https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu-v3/benchmarks/results/regressor-gpu.csv
However, I wouldn't mind about it at this point, as I'd focus on ensuring the correctness required to push into `main` for a release. 

Following a run on internal real-world data, the model predictions returned NaN on some instances; which doesn't occuer on cpu. 
An hyporthesis I had is that it may be tied to FLoat32 histogram accumulation. I've encountered different stability issues in the past, and currently on CPU, it uses FLoat64 for histograms (but FLoat32 for gradients and preds). I've created branch `grant-gpu-v4` that uses `Float64` for histograms. Test in underway, I'll keep you informed if it fixes the issue.
https://github.com/Evovest/EvoTrees.jl/tree/grant-gpu-v4
 
Another hypothesis is that the issue encountered may be tied to the anomaly observed in the `quantile` run: 
https://github.com/Evovest/EvoTrees.jl/blob/grant-gpu-v3/docs/src/assets/quantiles-sinus-binary-gpu.png
(the `cpu` version don't have the anomalies with predicitns over 1 or under 0. 
This may be just a red herring, but may be worth taking a look at the root cause for this. The change to Float64 hist didn't resolve this anomaly for quantiles.  

Would you mind syncing you branch with https://github.com/Evovest/EvoTrees.jl/pull/300? There's some formatting discrepancies (I'm following the default Julia's VSCode Formatter). 

In summary, I'm currently running a test on Float64, but there's likely a remaining small correctness issue that still need to be resolved.

AdityaPandeyCN:
Thankyou @jeremiedb for the updates, I worked on the quantile bug and think have resolved it. I will push this once I clean the code but I have some confusion regarding the subtraction trick since I have tried multiple times to remove it but it has resulted only in more regression in performance 
> Also, I'd be curious to test the performance if you skip the histogram subtraction, considering that currnetly you still iterate over the full list of `is`.

I agree with these but still I think that we are saving lot of time in histogram building. Maybe I can work on finding a solution of not iterating over the full` is` which can give us some more improvements. What do you feel about this?

jeremiedb:
> I had one more doubt like did we test our gpu-hist POC branch with the larger internal dataset? Currently I am thinking of ways to make it fast and doing a profiling. Any direction from you will be appreciated :). Also I think it would be good to have this code merged at this point if things look good to you and I can followback with a PR which makes more improvement?

Good point regarding robust approach to `gpu-hist` benchmarking. 
I'll look to adapt that `gpu-hist` branch to be able to validate that it results in in same fitting quality internally. That would ensure that the speed obserned in `regressor.jl` benchmarks are legitimate.

PR #300, opened 2025-09-16
Grant gpu v3

jeremiedb:
@AdityaPandeyCN Here I've open a PR  from the current state of your PR https://github.com/Evovest/EvoTrees.jl/pull/299, against the state of the PR a little over 2 weeks ago. I notived that the GPU performance has experienced some material slowdown. In benchmarks/results/regressor-gpu.csv, you can noticed how the  training time move up significantly, for example 1M, 100 feats, depth 11 moving from 7 sec to 14 sec. 

Would you have an idea what could have caused such regression? I noted that the gradient updates and evaluation metrics kernels were moved from CUDA to KA. 
Also, some changes to the `update_hist`. Is it necessary to introduce adhoc heuristics like https://github.com/Evovest/EvoTrees.jl/compare/grant-gpu-v2...grant-gpu-v3#diff-d696002f188318b01ef653ce29625f5597f4bc2f629f9885226494cc4286fd18R51?

Also, I noted that some `sunchronize()` were added. 

Let me know if you have some clear idea as where to look to fix this regression on training time. Otherwise, I could look at rerunning those benchmarks against an intermediary commit to help isolate where regression was introduced.

Also, from running basic test runs in https://github.com/Evovest/EvoTrees.jl/blob/99af060eb62e31f8a1aef7ccd1011e47b7f2b878/experiments/readme_plots.jl#L169, I noted that the `mae` fitted value diverges from the expected. value. Quantiles also seem to have an issue. I've seen that you create some branching logic to treat these losses functions. I'm a little worried about introducing such adhoc conditions as it exposes to harder maintenance. 

That being said, I wouldn't worry for now about the `mae` / `quantile` behavior, but rather isolate where the performance regression comes from. If it's a matter of using back some CUDA instead of KA, that's fine.


AdityaPandeyCN:
Hello @jeremiedb I feel the regression is coming from the excess synchronisation and kernel separation I am doing. I think I have made this really complicated I need to step back and make it more general rather than adding targeted code for loss functions as you pointed out.

AdityaPandeyCN:
Hello @jeremiedb Any updates on it? Is there any aspect which you would like me to work on?

AdityaPandeyCN:
Hello @jeremiedb I did some changes to address the problems. I ran the test file you provided and can see 99% correlation now. I did these changes like 

1. Aligned GPU categorical split logic with CPU : -evaluate all bins non-cumulatively in hist search and use single-bin sums in split application.
2. Fix node indexing: when a node doesn’t split, keep samples at the current node (avoid empty leaves/MAE fallback).

AdityaPandeyCN:
Yes, actually `update_hist_gpu_optimized! `only scans observations for build nodes (line 138 
 view(cache.build_nodes_gpu, 1:build_count_val)), not all active nodes. `subtract_hist_kernel!` then computes histograms for the remaining nodes via hist[child] = hist[parent] - hist[sibling], avoiding expensive observation scans. The bug is that after subtraction, I never call `find_best_split_from_hist_kernel! `for subtract nodes, leaving their split info uninitialized.

AdityaPandeyCN:
Hmm, I understand what are you implying. `hist_kernel!` already accumulates for all nodes based on nidx[obs], so calling it with only build nodes still processes observations for ALL nodes. I am working on it, will it be better to defer the internal dataset testing till then?

AdityaPandeyCN:
Hello @jeremiedb I have tried to address the problems, I removed the subtraction codes but it didn't go well and got regression in performance so have re implemented it keeping in mind everything you pointed out. Please have a look when time permits.

PR #301, opened 2025-10-13
Grant gpu v4

jeremiedb:


ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,330 @@
...
+    copyto!(tree.cond_bin, cache.tree_cond_bin_gpu)
+    copyto!(tree.gain, cache.tree_gain_gpu)
+    copyto!(tree.w, view(cache.nodes_sum_gpu, size(cache.nodes_sum_gpu, 1), 1:length(tree.w))) # FIXME: nodes_sum_gpu[k,:] should match tree.w size
+
+    leaf_nodes = findall(!, tree.split)
+
+    # MAE/Quantile: use CPU-based leaf prediction with exact observation indices
+    # Build leaf membership from nidx and compute predictions using actual sample quantiles
+    # instead of histogram-accumulated sums to avoid numerical errors.
+    if L <: Union{EvoTrees.MAE,EvoTrees.Quantile}
```
jeremiedb: Special loss treatment, where individual observations is needed, is specific  to `Quantile`. 
`MAE` should work following the generic approacah (even if it's not the regular gradient update rule, it only depends on aggregate metrics provided by the histograms).
See: https://github.com/Evovest/EvoTrees.jl/blob/d81c6033ab56264e02606264d008d84bf5e481e0/src/fit.jl#L132-L135

jeremiedb:
@AdityaPandeyCN While bringing to release prep adaptations, I noticed some potential low-hanging fruits for improvements and some rough edges. I think the above comments are worth adressing prior merging to `main`.

ext/EvoTreesCUDAExt/fit.jl
```diff
@@ -1,267 +1,337 @@
...
+        n_active = 1
+    end
+
+    # Main loop: build tree level by level - FIX: stop at max_depth-1
+    for depth in 1:(params.max_depth-1)
+        iszero(n_active) && break
+
+        view(cache.n_next_active_gpu, 1:1) .= 0
+        n_nodes = 2^(depth - 1)
+        active_nodes = view(cache.anodes_gpu, 1:n_nodes)
```
jeremiedb: Why not just set `active_nodes = view(cache.anodes_gpu, 1:n_active)`?
AFAIU, the functions `find_split_all!` or `apply_splits_kernel!` should only need the list of active nodes (as no op should happen on inactive ones.

PR #317, opened 2025-12-13
Parallelize GPU Split Finding Over (Node, Feature) Pairs

AdityaPandeyCN:
This PR improves GPU utilization during tree building by parallelizing split finding across both nodes and features simultaneously, rather than iterating over features sequentially within each node thread.
Changes
1.  New kernels in fit-utils.jl:

`compute_nodes_sum_kernel! `— Precomputes gradient sums for active nodes before split evaluation
`find_best_split_parallel_kernel!` — Evaluates splits with one thread per (node, feature) pair
`reduce_best_split_kernel! `— Reduces per-feature gains to find best split per node

2. Modified in fit.jl:

Updated root and depth≥2 split finding to use new three-kernel pattern

3. New cache buffers in init.jl and structs.jl:

`gains_per_feat_gpu` — Stores gain for each (feature, node) pair
`bins_per_feat_gpu` — Stores best bin for each (feature, node) pair
`sums_temp_par_gpu `— Per-thread accumulator for K>1 gradient sums

### DIFF
--- a/ext/EvoTreesCUDAExt/fit-utils.jl
+++ b/ext/EvoTreesCUDAExt/fit-utils.jl
@@ -252,12 +252,12 @@ Accumulate (atomic) gradient sums over observations `is` into the root node (nod
 end
 
 """
-	compute_nodes_sum_kernel!(nodes_sum, h∇, active_nodes, K)
+	compute_nodes_sum_kernel!(nodes_sum, h∇, active_nodes, js, K)
 
 Compute per-node gradient totals by summing histograms across bins.
 Writes into `nodes_sum[:, node]` for each node in `active_nodes`.
 """
-@kernel function compute_nodes_sum_kernel!(nodes_sum, @Const(h∇), @Const(active_nodes), K::Int)
+@kernel function compute_nodes_sum_kernel!(nodes_sum, @Const(h∇), @Const(active_nodes), @Const(js), K::Int)
     gidx = @index(Global)
     n_active = length(active_nodes)
     n_k = 2 * K + 1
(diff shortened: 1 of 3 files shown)


jeremiedb:
I could confirm the consistent results between the cpu and the new gpu runs. 
If you could take a quick shot at reviewing `find_best_split_parallel_kernel` complexity, that would be great. I'll then be looking integrating these improvements and release. Thanks for the nice work!

jeremiedb:
No issue not combining the `find` and `aggregate`; better not too force computation in same function if there's no a clear path to have the gpu "workers" efficiently assigned. 

I've just realized a subtle bug in the implementation: when `colsample < 1.0`, some trees end-up having no splits points, ending-up essentially as "dumb" trees. 
You can see how to repruce in the following file from a forked branch of this PR: https://github.com/Evovest/EvoTrees.jl/blob/jdb/gpu-perf/benchmarks/regressor-df.jl

You can reproduce by changing `colsample` from 1.0 to 0.5: 
https://github.com/Evovest/EvoTrees.jl/blob/5cc3b62bb494ef9ccdf3a0b78a2725e0cc9b2f7e/benchmarks/regressor-df.jl#L46

If `colsample` is set to 1.0, both CPU and GPU models are well aligned.
However, with `colsample = 0.5` there's almost half of the trees that have no splits. The other hald have a full number of trees (~ 31 in this case):
https://github.com/Evovest/EvoTrees.jl/blob/5cc3b62bb494ef9ccdf3a0b78a2725e0cc9b2f7e/benchmarks/regressor-df.jl#L77-L81

This isn't isn't observed in `main`, thus it must be some improper handling of the some feature-wise cache under this PR  parallel implementation. 

On the upside, when `colsample=1` and gpu reconcile with cpu results, the speedup from this PR  work remains strong.

PR #327, opened 2026-05-28
added multiquantile loss

### LINKED ISSUE #269: Feature Request MultiQuantile
It would be nice to have multi-quantile regression for approximating histogram in one go.
similar to
https://catboost.ai/en/docs/concepts/loss-functions-regression#MultiQuantile

benjamin-jsauve:
### Summary
Closes #269.

adds `:multiquantile` loss to `EvoTreeRegressor` for predicting K 
quantiles simultaneously in a single model.

### Changes
- New `MultiQuantile` loss type with gradient layout `[1:K]` , `[K+1:2K]` residuals, `[2K+1]` weights
- `EvoTreeRegressor` gains `alphas` parameter (default `[0.1, 0.5, 0.9]`)
- Exact residual quantiles for leaf values (same approach as single quantiel)
- MAE-style split gain summed across K outputs
- Empirical quantile bias initialization per alpha
- CPU + GPU support
- Tests and demo script

### Testing
```julia
Pkg.test("EvoTrees")
```

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -1,4 +1,4 @@
-struct CallBack{B,P,Y,C,D}
+struct CallBack{B,P,Y,C,D,A}
     feval::Function
     x_bin::B
     p::P
@@ -6,6 +6,7 @@ struct CallBack{B,P,Y,C,D}
     w::C
     eval::C
     feattypes::D
+    alphas::A
 end
 
 function CallBack(
@@ -22,7 +23,6 @@ function CallBack(
     _offset_name = isnothing(offset_name) ? Symbol("") : Symbol(offset_name)
     _target_name = Symbol(target_name)
 
-    feval = metric_dict[params.metric]
     x_bin = binarize(deval; feature_names=m.info[:feature_names], edges=m.info[:edges])
     nobs = length(Tables.getcolumn(deval, 1))
     p = zeros(T, K, nobs)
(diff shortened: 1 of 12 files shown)


src/callback.jl
```diff
@@ -43,8 +43,11 @@ function CallBack(
     else
         y = T.(y_eval)
     end
+    feval = metric_dict[params.metric]
     V = device_array_type(device)
     w = isnothing(weight_name) ? device_ones(device, T, length(y)) : V{T}(Tables.getcolumn(deval, _weight_name))
+    alphas_eval = hasproperty(params, :alphas) ? T.(params.alphas) : nothing
+    device <: GPU && !isnothing(alphas_eval) && (alphas_eval = V{T}(alphas_eval))
```
jeremiedb: Not a blocker; but it would be good to think of a pattern that avoids introducing loss-specific parameters into structures intended to act as loss-agnostic, generic components. Ie: the interface for `CallBack` / `eval metric` would ideally not be impacted by the introduction of a new loss / eval-metric.

src/learners.jl
```diff
@@ -476,6 +480,21 @@ function check_parameter(::Type{<:T}, value, min_value::Real, max_value::Real, l
     end
 end
 
+function check_alphas(alphas)
+    try
+        @assert length(alphas) >= 2
```
jeremiedb: Any objection to accept a length(alphas) = 1?
I'm thinking forward that the `quantile`/`multiquantile` could be merged into a single loss for simplicity.

src/learners.jl
```diff
@@ -476,6 +480,21 @@ function check_parameter(::Type{<:T}, value, min_value::Real, max_value::Real, l
...
+        prev = -Inf
+        for alpha in alphas
+            a = Float64(alpha)
+            @assert 0.0 < a < 1.0
+            @assert a > prev
+            prev = a
+        end
+    catch
```
jeremiedb: consider `issorted` / `extrema` for simplification

jeremiedb (commented):
Thanks for the PR! 
Looks good and functional. I'd just have minor simplification comments. Ready to merge then.

benjamin-jsauve:
For 1, I replaced the loss-specific cb.alphas with a generic metric_kwargs NamedTuple so callbacks are loss‑agnostic and call feval(...; cb.metric_kwargs...).

For 2 and 3, exactly as you described.

Thanks for the comments!

jeremiedb (approved):
Thanks for the PR!

PR #328, opened 2026-06-13
Add multi-target regression (:mtmse): joint vector-target fitting on CPU + GPU

AdityaPandeyCN:
Add multi-target regression (`:mtmse`): joint vector-target fitting on CPU + GPU.
Fits one tree model to multiple targets jointly with shared structure, vector leaf, and split gain summed over targets. Use via `loss=:mtmse` with `target_name=[:y1, :y2, :y3]` (or a `(K, nobs)` matrix); predictions return `(nobs, K)`.
 
Implemented as K = n_targets stacked into the existing `2K+1` histogram axis with a single shared weight row, not a separate 5D dimension. Reuses the already-K-generic histogram and split kernels unchanged and stores weight once

### DIFF
--- a/blog/cred/utils.jl
+++ b/blog/cred/utils.jl
@@ -1,6 +1,6 @@
 ## package loading
 using EvoTrees
-using EvoTrees: _get_cred, _loss2type_dict, update_grads!, Cred, EvoTypes, GradientRegression
+using EvoTrees: _cred_Z, _loss2type_dict, update_grads!, Cred, EvoTypes, GradientRegression
 using DataFrames
 using Distributions
 using Statistics: mean, std
@@ -32,7 +32,8 @@ function get_simul_metric(metric_name="cred"; nobs, loss, spread=1.0, sd=1.0)
     ∑ = get_∑(p, y, L, config)
 
     if metric_name == "cred"
-        metric = _get_cred(L, config, ∑)
+        ϵ = eps(eltype(∑))
+        metric = _cred_Z(L, ∑[1], ∑[2], ∑[end], ϵ)
     elseif metric_name == "gain"
         metric = _get_gain(L, config, ∑)
     else
@@ -68,7 +69,8 @@ function get_data(; loss, nobs, spread=1.0, sd=1.0)
     data[:gC] = data[:gL] + data[:gR]
 
     if L <: Cred
-        data[:ZR] = _get_cred(L, config, ∑R)
+        ϵ = eps(eltype(∑R))
+        data[:ZR] = _cred_Z(L, ∑R[1], ∑R[2], ∑R[end], ϵ)
     else
         data[:ZR] = NaN
     end
--- a/ext/EvoTreesCUDAExt/fit-utils.jl
+++ b/ext/EvoTreesCUDAExt/fit-utils.jl
@@ -309,11 +309,12 @@ end
 
 # Parent gain: MLE2P
 @inline function parent_gain(::Type{L}, nodes_sum, node, K, λw, L2, w_p, ε::T) where {T,L<:EvoTrees.MLE2P}
-    g1, g2 = nodes_sum[1, node], nodes_sum[2, node]
-    h1, h2 = nodes_sum[3, node], nodes_sum[4, node]
-    d1 = max(h1 + λw + L2, ε)
-    d2 = max(h2 + λw + L2, ε)
-    return (g1^2 / d1 + g2^2 / d2) / 2
+    gain = zero(T)
+    for k in 1:K
+        g, h = nodes_sum[k, node], nodes_sum[K+k, node]
+        gain += g^2 / max(h + λw + L2, ε)
+    end
+    return gain / 2
 end
 
 # Parent gain: MLogLoss
--- a/ext/EvoTreesCUDAExt/fit-utils.jl
+++ b/ext/EvoTreesCUDAExt/fit-utils.jl
@@ -442,6 +442,23 @@ end
 @inline function split_gain_multi(
     ::Type{L}, sums_temp, nodes_sum, node, temp_idx,
     K, w_l, w_r, gain_p, lambda, L2, ε::T
+) where {T,L<:Union{EvoTrees.MAE,EvoTrees.Quantile}}
+    w_p = nodes_sum[2 * K + 1, node]
+    d_l = max(1 + lambda + L2 / w_l, ε)
+    d_r = max(1 + lambda + L2 / w_r, ε)
+    gain = zero(T)
+    for k in 1:K
+        g_l = sums_temp[k, temp_idx]
+        g_r = nodes_sum[k, node] - g_l
+        gain += abs(g_l / w_l - nodes_sum[k, node] / w_p) * w_l / d_l
+        gain += abs(g_r / w_r - nodes_sum[k, node] / w_p) * w_r / d_r
+    end
+    return gain - gain_p
+end
+
+@inline function split_gain_multi(
+    ::Type{L}, sums_temp, nodes_sum, node, temp_idx,
+    K, w_l, w_r, gain_p, lambda, L2, ε::T
 ) where {T,L}
     return T(-Inf)
 end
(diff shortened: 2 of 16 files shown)


jeremiedb:
@AdityaPandeyCN Just as an initial pointer; in order for the implementation to scale smoothly I see as important for the multi-target support to not come with its own new set of loss/metric functions; but instead reuse the actual mechanisms already in place fo rthe single target loss; but generalize then so they can be applied over a vevtor of targets rather than a scalar.

Ie: for a given `learner` (ex: `EvoTreeRegressor(loss="logloss", ...); one coud call fit with either target_name = "y" or `target_name = ["y1", "y2"]. The same tree building mechanism would apply, except that the gradients would need to be computed and accumulated over a new length Y dimension in order to compute each of the target prediction. 
Going through individual `mse-multi` / `logloss-multi` will likely end-up in much of duplications and more maintenance burden, than a single principle that general to any single target loss function, resulting in support for multi-target support for any loss function for "free"

AdityaPandeyCN:
One scope decision I'd like your call on. The single-parameter losses generalize cleanly: each target contributes one gradient + one hessian, stacking into the existing 2K+1 layout with K = n_targets. The multi-parameter losses are different , like gaussian_mle/logistic_mle use their output dimension for the distribution's two parameters (μ, logσ), and softmax/multiquantile similarly. 

Do we plan to work on multi-paramter losses too?

jeremiedb:
@AdityaPandeyCN Yes I want multi-parameter losses to be supported as well. At least, the `EvoTreeMLE`. 
Under the current implementation, K can be understood as the number of parameters for a given loss functions, resulting in the 2 * K + 1 information elements that are tracked in the per-observations gradients and the histograms.

As I understand, you're initial approach has been to use that `K` as refering to the number of targets, which works fine for single parameter losses (mse, logloss). It however result in potential ambiguities for GaussianMLE or other multi-parameter as you pointed out. 

In order to maintain a clear code design, I'm thinking that it would be preferable to distinguish the terms used to refer to the number of parameters for a given loss and the number of targets.
So the number of elements could be:
- `K * Y + 1`, where K=number of elements per loss (K=2 for mse, 4 for gaussianMLE, etc.).

Regarding the shape of the object, open options (for the gradient object):
- `[K * Y + 1, nobs]` stack all the info elements in the same dimension
- `[K + 1, Y, nobs]`  target has its own dimension. It may be inefficient as it would result in repeated `weight` (the + 1 element)

My current take from the above is that the approach of storing all of the target information in the gradient/histogram along the same dimension see to make sense. A potential caveat to think about is whether the handling of the gains and the the leaf predicitons can smoothly generalize when computed over this stacked information. Ie: likely need to loop over the Y target on the `K * Y + 1` info; vs have a vectorized approach under the `[K + 1, Y, nobs]` layout. 

Let me know what you think.

AdityaPandeyCN:
Hello,
On your caveat about gains and leaf predictions: they generalize by looping over the Y targets. For `GaussianMLE` the gradient is `4Y+1` (2Y grads, 2Y hessians, one shared weight); `get_gain` and the leaf prediction loop over each target's parameter block, so no weight repetition and the `[K+1, Y, nobs]` layout wasn't needed.
K (params per loss) and Y (targets) are kept distinct  like single-target reduces to the existing code path, multi-target uses `K*Y+1`.

In the latest commit have tried to add multi-target EvoTreeMLE (gaussian_mle, logistic_mle) on the stacked `[K*Y+1, nobs]` layout.
`update_grads!`, `get_gain`, and leaf prediction loop over the Y targets (each a 2-param μ/logσ block); `init` sets K=2Y for matrix targets. Removed the hardcoded 5-row MLE2P histogram so it generalizes over the stacked dimension, and generalized `GPU parent_gain` the same way. Single-target reduces to the existing path.

Please take a look.

jeremiedb:
Looks good!
A small caveat: the eval/metric function for `gaussian_mle` loss seems not adapted for multi-target. 
See the following test where using `deval` errors: 
https://github.com/Evovest/EvoTrees.jl/blob/b1f11b3978bbfdfe14780478af3bf437e5bf4261/experiments/readme_plots-multi-target.jl#L133

The `deval` also result in errors when executingb on `gpu` (scalar operation). Can be reproducved from https://github.com/Evovest/EvoTrees.jl/blob/jdb/multi-target/experiments/readme_plots-multi-target.jl

AdityaPandeyCN:
@jeremiedb I have tried to address the concerns.  Please take a look when time permits.

jeremiedb:
@AdityaPandeyCN: running for other loss functions still fail as the `metrics` for other than mse/gaussian_mle haven't been adapted to support multi-target. See https://github.com/Evovest/EvoTrees.jl/blob/jdb/multi-target/experiments/readme_plots-multi-target.jl where running through the `loss_list` fails for losses other than `mse`. 

Do you see any blocker for have complete loss/metric coverage for multi-target? I think the classifier (`mlogloss`) could be ignored though, but I'd otherwise expect full support.

AdityaPandeyCN:
Hello, added support for the loss functions from the `loss_list`, what will be the move for the `mae` and `quantile` as they wont be generic additions?

jeremiedb:
What would be the issue if following the same approach as for other losses and defining multi-target  (`_mt_`) version of the various kernels for these losses? 

Regarding the general design approach; the codebase for the losses and metrics seems to have substantially inflated. Could a more generic approach have worked where the each of the loss/metric definition at the single observation/target is defined; and then generic functions apply these core kernels to each observations and target?

AdityaPandeyCN:
Hi, have made generalisation of the loss and metric functions, how do you wish to test it apart from the readme plots file? Let me know so that I test it completely from my end before a review from your side.

jeremiedb:
The adaptations look clean!
From the running the multi-target figures readme: https://github.com/Evovest/EvoTrees.jl/blob/jdb/multi-target/experiments/readme_plots-multi-target.jl
it looks like the only unsupported regression tasks would be `cred_var`, `cred_std` and `multiquantile`. 
I'm fine not supporting `multi-quantile`, but I'd expect that `cred_var` and `cred_std` should be straightforward. 

As for testing, I'd suggest to add a distinct test file for multi-target, that would essentially perform what's done in the above multi-target plots generation: loop over all supported losses and have apply a basic check on the resulting MSE as a sanity check.

AdityaPandeyCN:
Hello, added support for `cred_var` and `cred_std`, please take a look.

I had a question regarding the harmonization objective. With all the analysis I did, I feel the best way would be to complete the KernelAbstractions.jl migration, and once things are generalised with KA.jl, investigating CPU performance would be easier.
Also, since there is active development going on for the OpenCL.jl project, we could also add support for this via PoCL. 
Let me know what would be better according to you.

jeremiedb:
I'm aligned that completing the KernelAbstractions story would be a beneficial first step. From what I recall, it would mostly covers the loss and eval/metric functions, so should normally be a fairly smooth path. And there were some initial allocations/fill operations that relied on explicitly on CUDA I tihnk.

I'm not much familiar with PoCL; is that the mechanism through which KernelAbstractions intend to support CPU backend?
It would definitly be an interesting outcome if simply using a CPU target from KernelAbstraction resulted in proper CPU performance; de facto unifying the codebase. 

However, I'd be reluctant in enforcing a strict dependency on KA/OpenCL for CPU execution as I'd ideally like to get vanilla Julia code achieve good performance. 

So I'm aligned on first completing KA story (thus activating ability to execute on ROCM or Metal) and check if OpenCL can work effectively. Then, assess where the performance story lands and identify the chokepoints in CPU performance.

AdityaPandeyCN:
> I'm not much familiar with PoCL; is that the mechanism through which KernelAbstractions intend to support CPU backend? It would definitly be an interesting outcome if simply using a CPU target from KernelAbstraction resulted in proper CPU performance; de facto unifying the codebase.

 I think this would be the backend to support running SPMD(Single Program, Multiple Data) programme, different from pure cpu backend.

> However, I'd be reluctant in enforcing a strict dependency on KA/OpenCL for CPU execution as I'd ideally like to get vanilla Julia code achieve good performance.

Yes, not making this default but setting up a ground so it can be easily test in future or once it is stable.

Issue #329, opened 2026-07-14
Feature Request: early stopping round tolerance


tecosaur:
Hello,

I was tuning a classifier tree for a problem, and I found that I was always hitting the maximum number of rounds, even though for the latter half the improvement was negligible.

Looking at the per-iteration performance, there's a point where iterations don't meaningfully improve, but the strict inequality here:

https://github.com/Evovest/EvoTrees.jl/blob/1e6c7a79b8d18124545f20737f1a0fef3fddf03d/src/callback.jl#L148-L149

means that an insignificant improvement is counted. It would be useful if I could set a tollerance term, like `early_stopping_tolerance=1e-3`.

From a quick survey, it looks like this is a common knob in other GBDT libraries (XGB, LGBM, CatBoost, scikit).

jeremiedb:
@tecosaur For curiosity, since I've not been exposed to much cases where the out-of-sample metric don't end-up deteriorating, could you share some details about the context of this classification problem?
Ex:, number of bservations, features, class imbalance, etc? Also, is the model capacity large (max_depth, nbins)?
I'm asking as typically, GBT models overfit and exhibit gaps between training and evaluation data, such that out-of-sample end up having a clear deterioration rather than random oscillation. 
What was the `nrounds`?

tecosaur:
Sure! Here's the situation

- Binary classification task with `EvoTreesClassifier`
- 61 features
- 340k rows/observations
- ~3% positive rate
- Max depth 6
- Bagging size 12
- nrounds 2000
- nbins 224
- eta 0.10

The rows are comprised of groups of ~12 (1-25) within which only one feature changes. Since I can't make the subsampling take this into account (this is the situation described in #287), I expect this leakage into the out-of-sample data causes the distributions to be very similar, and so unable to effectively assess when overfitting is happening.

I've tried to address this by using a group-aware splitting of the data into five folds, and then training five classifiers on 4/5 folds and evaluating the final model on the out-of-fold data (as you recommended).

jeremiedb:
Thanks for the info. Have you seen bagging size had a substantial benefit? When having a small enough learning-rate leading to a large number of trees (a few hundred or more, like in you situation), I haven't observed yet a compelling benefit from this hyper-paramater. 
Had you performed hyper-search, notably with greater depth? I can be worth testsing with up to the 10-12 range. Maybe there were some represenattion capacity involved. 
[...]

PR #331, opened 2026-07-14
Add early_stopping_tolerance parameter for early stopping

AdityaPandeyCN:
Adds an `early_stopping_tolerance` parameter so that improvements smaller than the tolerance no longer reset the early-stopping counter.
PR to address #329

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -123,11 +123,12 @@ function (cb::CallBack)(logger, iter, tree)
     return nothing
 end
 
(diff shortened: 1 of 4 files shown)


jeremiedb:
Thanks, looks good!
Could you just add a test that ensures that the feature is effectively functional (can be limited to a single `mse` regression task on a dummy data).

jeremiedb (approved):

PR #332, opened 2026-07-21
Complete KernelAbstractions migration for portable GPU backends

AdityaPandeyCN:
Moves all GPU kernel logic out of the CUDA-specific extension into a single backend-neutral `EvoTreesKernelAbstractionsExt,` rewritten against `KernelAbstractions.jl.`


AdityaPandeyCN:
@jeremiedb Although the PR is still not perfect , how do we plan to test with different backends(AMD and metal)?

AdityaPandeyCN:
@jeremiedb Did you have any time to test the PR?

jeremiedb:
Sorry I've had a few conflicting priorities. I should take a look  at it tomorrow. 

I won't be able to test on AMD nor Metal in the short term. However, if you manage to have the the execution to run the on the openCL/cpu backend of KA; then I think it should provide a reasonable demonstration of the multi-backend compatibility?

I wouldn't be too concerned about proper AMD/Metal at the moment; as the primary justification for it was a potential low-hanging fruit for improved CPU performance. The original scope was to assess the potential for aligning the CPU algos with the 4D histogram layout; with the ultimate goal of limiting the overhead from large tree depth. 

Did you run some benchmarks.jl to validate that GPU/CUDA speed wasn't hit by the changes?

AdityaPandeyCN:
> I won't be able to test on AMD nor Metal in the short term. However, if you manage to have the the execution to run the on the openCL/cpu backend of KA; then I think it should provide a reasonable demonstration of the multi-backend compatibility?
>
I was able to run it on OpenCL, but had to patch around a float atomics gap so I didn't push it. 
Also hit `GPUArrays` logical indexing in `subsample`.
 
> Did you run some benchmarks.jl to validate that GPU/CUDA speed wasn't hit by the changes?

Ran the regressor.jl and the performance looks similar for CUDA backend

jeremiedb:
Closing in favor of #340

PR #334, opened 2026-07-30
complete kernelabstraction migration

jeremiedb:


jeremiedb:
@AdityaPandeyCN Can you check the failed CI on 1.10:
```julia
  ┌ Info: initialization
  └   metric = 0.12850541855989803
  EvoTreeRegressor - MSE: Error During Test at /home/runner/work/EvoTrees.jl/EvoTrees.jl/test/core.jl:25
    Got exception outside of a @test
    TaskFailedException
    
        nested task error: llvmcall only supports intrinsic calls
        Stacktrace:
         [1] _prefetch
           @ ~/work/EvoTrees.jl/EvoTrees.jl/src/fit-utils.jl:340 [inlined]
```

It appears tied to the direct LLVM call:
https://github.com/Evovest/EvoTrees.jl/blob/cdd6632a95034169a07d4e1b921f5b35ac7aa007/src/fit-utils.jl#L334-L342
I tihnk a more robust approach would be warranted here.

PR #340, opened 2026-08-02
Harmonize CPU/GPU implementation

AdityaPandeyCN:
I aim to achieve harmonization between CPU and GPU implementations with this PR. It needs https://github.com/Evovest/EvoTrees.jl/pull/332 to be merged.

Commit wise information of the PR are:-
a119531 Share loss and metric scalar math between CPU and GPU backends.

- Moved backend-agnostic loss/metric helpers into src/
- KA kernels call that single definition instead of duplicated formulas
- Stops CPU/GPU numeric drift

43cba18 Extract shared _init_target for CPU and GPU initialization.

- One path for target checks and deriving K, y, and bias
- CPU init_core calls it directly
- GPU init calls it on host, then copies arrays to device

f0160af Store CPU histograms in shared 4D arrays and unify subtract_hist!.

- GPU already used 4D h∇ / h∇L / h∇R as (grad × nbins × features × nodes)
- CPU now uses the same layout so both backends share cache shape and node-indexed views for split search
- Sibling hist via shared subtract_hist!: parent - other with a reshaped contiguous kernel (no aliasing broadcast)

c20cb8c Add observation-major histogram build with software prefetch.

- Keep transposed `x_bin_T `so a row's bins are contiguous
- Depth 1: feature-major over` x_bin` (thread over features)
- Depth >= 2 (K=1): obs-major over `x_bin_T `with software prefetch ahead (XGBoost RowsWise-style)


jeremiedb:
Thanks, this is an intereating development!

CPU benchmarks, while many are in parity zone, and 10M / depth 11 / feats 100 even showing an improvement, there are some regression of concern, mostly the 1M / 100 feats / depth 6; that almost doubled: 
https://github.com/Evovest/EvoTrees.jl/compare/jdb/harmonisation?expand=1#diff-dcb93c03cf96f9dc998db7262fcf7a65d9e514fbb2a8febc7f70c58926924604


- could you provide further rational for `llvm.prefetch`? Does it have an important effect on the performance? I'm looking to assess the trade-off between performance and simplicity / ease of maintenance by avoiding low-level tricks. Any equivalent in regular Julia that could achieve most of its benefit?

- Similarly, does the custom `_hist_obs_major` that's only used in first pass also adds substantial benefit compared to feat-wise only parallelism?

- Was it considered to adopt the node-index GPU approach for the CPU version? Especially for larger depth, I suspect this to be a potentially substantial perf bottleneck for CPU, since explicit "left/righ" vector of row-idx need to be built instead of just mutating a simple "idx" vector indicating to which node each observations is now mapped to. Such idx approach may however impair the performance of the downstream histograms. 
- CPU: https://github.com/Evovest/EvoTrees.jl/blob/a9eabe5e1a6953fce062c59e8f5bfad013d9d8b5/src/fit-utils.jl#L93
-  GPU: https://github.com/Evovest/EvoTrees.jl/blob/a9eabe5e1a6953fce062c59e8f5bfad013d9d8b5/ext/EvoTreesKernelAbstractionsExt/fit-utils.jl#L9

My overall appreciation is that this PR seems to go in a opsitive direction as it provides improved consistency across cpu-gpu. I'd like to have some further performance investigation performed to address some of the above concerns. 

Just a random thought: I experimented some time ago with static vectors for the first dimention of histograms (ex 3D array of Static vectors of length 2K+1). It show good performance on the hist accumulation, but would introduce differences with GPU approach, so may not be a desirable path.

AdityaPandeyCN:
> * could you provide further rational for `llvm.prefetch`? Does it have an important effect on the performance? I'm looking to assess the trade-off between performance and simplicity / ease of maintenance by avoiding low-level tricks. Any equivalent in regular Julia that could achieve most of its benefit?

I am not sure but actively looking if there is any in Julia. I got this idea from `xgboost` where they have done something similar (https://github.com/dmlc/xgboost/blob/master/src/common/hist_util.cc#L353-L365). On my local run it provided really good improvements but confused that it only improved the last depth for you. 

> * Similarly, does the custom `_hist_obs_major` that's only used in first pass also adds substantial benefit compared to feat-wise only parallelism?


> * Was it considered to adopt the node-index GPU approach for the CPU version? Especially for larger depth, I suspect this to be a potentially substantial perf bottleneck for CPU, since explicit "left/righ" vector of row-idx need to be built instead of just mutating a simple "idx" vector indicating to which node each observations is now mapped to. Such idx approach may however impair the performance of the downstream histograms.

This was a major part of the initial analysis and somehow using the gpu idea for histogram building caused regressions in my testing. But splitting is different than histogram building, I will see how to go about it.

P.S.-- Actually found something julia native https://github.com/JuliaSIMD/LoopVectorization.jl/blob/main/src/codegen/lower_load.jl
but this is not documented and will add an extra dependency.

AdityaPandeyCN:
@jeremiedb Can you test the benchmarks once again? 

I suspect thread count explains the gap: obs-major hist is serial per node and only turns on when `length(build_nodes) >= nthreads(),` so at max_depth=6 (few build nodes) a 12-thread machine can stay on the slower path / under-use cores, while my runs at threads=4 still hit the gate and showed gains.  Apologies for the inconvenience

AdityaPandeyCN:
I tried a new design closer to the GPU hist schedule: for single-target losses (K == 1), build `obs-major` as a (`node × row_block`) grid. Each task owns one (node, row_block), writes into its own temporary hist (`h∇_tls)`, then those are added into the real `nodes[n].h`. So a few large nodes can still use many cores, instead of obs-major being serial per node and gated on `length(build_nodes) >= nthreads()` .

GPU `nidx` type approach was still slower in my testing and this was closer to what I can do to bring GPU/CPU consistency.

jeremiedb:
Latest changes brought nice speed improvement for 10M obs, 100 features runs!
https://github.com/Evovest/EvoTrees.jl/pull/341/changes#diff-dcb93c03cf96f9dc998db7262fcf7a65d9e514fbb2a8febc7f70c58926924604R1-R13

Were you seeing other items pending to complete the scope of the grant? 
If reasonable effort, would you consider looking at fixing the support for Oblivious trees on GPU? https://github.com/Evovest/EvoTrees.jl/blob/6247b5ff3d185717a3e3730184d3beb3f37000c6/ext/EvoTreesCUDAExt/fit.jl#L22-L41

AdityaPandeyCN:
Sure, looking into it. Let me know if there are other aspects which you want me to address.

AdityaPandeyCN:
Can you take a look?

src/loss.jl
```diff
@@ -33,8 +33,21 @@ const _loss2type_dict = Dict(
...
-@inline _target(y::AbstractMatrix, k, i) = y[k, i]
+################################################################################
+# Backend-agnostic scalar math.
+#
+# Everything below is scalars-in / scalars-out, allocation-free and `@inline`d.
+# It is called both by the threaded CPU `update_grads!` loops in this file and
+# by the KernelAbstractions kernels in `ext/EvoTreesKernelAbstractionsExt/`, so
+# that each loss expression has exactly one definition in the package.
+#
+# `@propagate_inbounds` (rather than plain `@inline`) on the accessors is load
```
jeremiedb, line 44: Couldyou clarify a little further the rationale here? 
I'm not clear to understand the specific role of propagate inbounds here, and how defining just a naive function would impair performance?

jeremiedb:
@AdityaPandeyCN COuld you confirm which of the follwoing branch has the proper version:
- https://github.com/Evovest/EvoTrees.jl/pull/341
- https://github.com/Evovest/EvoTrees.jl/pull/342

In 341, is the original I intended to keep sync with https://github.com/AdityaPandeyCN/EvoTrees.jl/tree/harmonisation
However, it looks like it diverges and in 342, there appears to be missing some changes incorporated in 341, essentially from this commit: https://github.com/Evovest/EvoTrees.jl/pull/341/changes/7334f6f52fa1752fc0abd62c7fd42b023c026c18
See for example the following added config in `structs.jl` now absent in you branch `harmonisation`: https://github.com/Evovest/EvoTrees.jl/blob/8e1de8517214ad671bb261c57b48e80f3332c6c2/src/structs.jl#L19-L25

Please confirm whether branch are up to date, and I'll then go over final benchmarking, incl. on oblivious.

AdityaPandeyCN:
@jeremiedb sorry I messed up with the commits on this PR. This PR has all the changes https://github.com/Evovest/EvoTrees.jl/pull/342. I have pushed the missing commit here also including the refactor commit addressing your comment so you can pull them to https://github.com/Evovest/EvoTrees.jl/pull/342

jeremiedb:
I got an error running on GPU:
```julia
ERROR: LoadError: MethodError: no method matching _select_binary_split!(::EvoTreesKernelAbstractionsExt.CacheBaseGPU{CuArray{Float32, 1, CUDACore.DeviceMemory}, EvoTrees.TrainNode{SubArray{UInt32, 1, Vector{UInt32}, Tuple{UnitRange{Int64}}, true}, Vector{Float64}, Matrix{Float64}, Array{Float64, 3}}}, ::CUDABackend, ::Type{EvoTrees.MSE}, ::EvoTreeRegressor, ::CuArray{Int32, 1, CUDACore.DeviceMemory}, ::Int64, ::Int32)
The function `_select_binary_split!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _select_binary_split!(::EvoTrees.CacheGPU, ::Any, ::Type{L}, ::Union{EvoTreeClassifier, EvoTreeCount, EvoTreeGaussian, EvoTreeMLE, EvoTreeRegressor}, ::Any, ::Int64, ::Int64) where L
   @ EvoTreesKernelAbstractionsExt ~/evovest/EvoTrees.jl/ext/EvoTreesKernelAbstractionsExt/fit.jl:47
```

Looks like an inconsistency use of Int32 vs Int64.

AdityaPandeyCN:
Apologies, have fixed it now.

jeremiedb:
@AdityaPandeyCN To follow-up on the comment I made in `jdb/harm-v2` regarding performance regression on gaussian-MLE, I've rerun the benchmark on current `main`, `jdb/KA` and `jdb/harm-v2` (sync with this `harmonization` PR):

- https://github.com/Evovest/EvoTrees.jl/blob/main-bench/benchmarks/results/regressor-gaussian_mle-cpu.csv
- https://github.com/Evovest/EvoTrees.jl/blob/jdb/KA/benchmarks/results/regressor-gaussian_mle-cpu.csv
- https://github.com/Evovest/EvoTrees.jl/blob/jdb/harm-v2/benchmarks/results/regressor-gaussian_mle-cpu.csv

As can be seen, while `KA` remains mostly aligned with `main`, the regression appears specific to change from this `harmonisation` update.

AdityaPandeyCN:
@jeremiedb I have attempted some more harmonisation, Can you run the benchmarks from your side?

jeremiedb:
@AdityaPandeyCN I see some performance regression on CPU for the 10M obs. case, but I'd consider it acceptable considering the benefits from improved CPU/GPU integration. 

I'd mve with mergin into this branch and then the `jdb/KA` for final bencharmking prior to merge on main. 
In the meantime, could provide a quick summary of key design changes and learnings from it? 

For example, the drop in hL/hR is a material change; as I understand, it could be dropped in favor a scan over bins that only required to track scalers instead of full histogram pre-allocations. Is this correct?

src/loss.jl
jeremiedb: @AdityaPandeyCN Can you comment on this?

AdityaPandeyCN: Sorry, missed this.
The old code had these accesses written inline inside `@inbounds` blocks, so no bounds checks. Per the docs, `@inbounds` doesn't elide checks when the access sits inside a function call, so moving them into `_target` would have reintroduced a check per access in the hot loops. `@propagate_inbounds` keeps the extracted function check-free inside `@inbounds` callers.
https://docs.julialang.org/en/v1/devdocs/boundscheck/#Eliding-bounds-checks

AdityaPandeyCN:
Yes, that's right.

CPU used to `cumsum!` the node hist into `hL`/`hR` (two extra 4D buffers, same size as the hist). Split scan now keeps a running left (`∑L`) and sets right as parent minus left. `∑R` is only filled when a monotone constraint is on. GPU was already on that convention.

For `K == 1` the left state is three numbers (`g`, `h`, `w`). For `K > 1` it is a `2K+1` vector, not a full `[2K+1, nbins, feats]` hist.

Other design notes:

* CPU hist is the same 4D layout as GPU: `h∇[2K+1, nbins, feats, nodes]`. Subtract is `parent - sibling` on that array.
* Build is one obs-major kernel (`x_bin_T` + prefetch), threaded over `(node × feature-tile)`. Tiles own disjoint `h[:, :, j]`, so no atomics and no thread-local hist slabs. Same kernel for MSE and Gaussian (Gaussian used to sit on the slow `K != 1` path).
* Split formulas live in `src/split.jl` (`node_gain` / `split_gain`). CPU and GPU call the same math.
* We did not put GPU `nidx` on CPU. Tried; it was slower. CPU still uses `is` lists + `split_set!`.
* GPU `K == 1` stays in registers. Monotone checks were left per-backend (`pred_scalar` vs `check_monotone`).

I am still understanding how scaling more no of threads are not helping the performance improvements because it does for threads=4.

ext/EvoTreesKernelAbstractionsExt/fit-utils.jl
```diff
@@ -492,147 +300,83 @@ end
...
+            nodes_sum[1, node] - acc1, nodes_sum[2, node] - acc2,
+            w_l, w_r, lambda, L2, ε,
+        ) && return (T(-Inf), acc1, acc2, accw)
+        ∑ = (nodes_sum[1, node], nodes_sum[2, node], nodes_sum[3, node])
+        ∑L = (acc1, acc2, accw)
+        gain = EvoTrees.split_gain(L, ∑, ∑L, w_l, w_r, lambda, L2, ε) - gain_p
+        return (gain, acc1, acc2, accw)
     else
-        accumulate_hist_kn!(sums_temp, h∇, f, b, node, K, is_numeric, temp_idx)
+        EvoTrees._acc_left!(sums_temp, temp_idx, h∇, f, b, node, 2 * K + 1, is_numeric)
```
AdityaPandeyCN, line 357: I look forward to eliminate this branching too, but only if we retain the performance.

AdityaPandeyCN:
> @AdityaPandeyCN I see some performance regression on CPU for the 10M obs. case.

Can I know the numbers? Like curious if it regressed for `mse `and `gaussian_mle` both

jeremiedb:
For benchmarks, following latest state, the gap appear very thin:

- https://github.com/Evovest/EvoTrees.jl/blob/jdb/harm-v2/benchmarks/results/regressor-mse-cpu.csv
- https://github.com/Evovest/EvoTrees.jl/blob/main-bench/benchmarks/results/regressor-mse-cpu.csv

Ex: 1M/100 feats/depth-6:  from 5.1s to 6.2s.
I consider the general portrait favourable. If you can think of further improvements without substantial code complexity, it will naturally be welcome; but I don't see this as a bottlneck for now.

PR #341, opened 2026-08-06
Jdb/harmonisation

jeremiedb:


src/fit.jl
```diff
@@ -150,15 +153,20 @@ end
...
     js,
     is,
     left,
     right,
     x_bin,
+    x_bin_T,
     feattypes::Vector{Bool},
-    monotone_constraints
+    monotone_constraints,
+    h∇_tls,
```
jeremiedb, line 169: Could you document the logic for the `tls` block? In particular, add detail to their usage/dependency?

src/fit-utils.jl
```diff
@@ -312,70 +312,157 @@ function split_set_single!(
...
-function update_hist!(
-    ::Type{L},
-    hist::Array,
-    ∇::Matrix,
-    x_bin::Matrix,
-    is::AbstractVector,
-    js::AbstractVector,
-) where {L}
-    hist .= 0
+function update_hist!(nodes, build_nodes, ∇, x_bin, x_bin_T, js, h∇_tls)
```
jeremiedb, line 356: Could a dispatch mechanism may have been preferred here?
My general overall impression is that which block gets applied in a given context is not that intuitive here. 

For example, could the dispatch logic for each of the 3 update hist optiosn? (1 `_hist_feat!` vs 2 `_hist_obs!`)

jeremiedb:
Closing in favor of #342

PR #342, opened 2026-08-09
Jdb/harm v2

jeremiedb:


AdityaPandeyCN:
@jeremiedb Did it run with GPU for you? Also please let me know if I need to make the docstrings more informative.
I tried adding more things but afraid that it will make the modules more bloated.

jeremiedb:
@AdityaPandeyCN While the CPU benchmarks look fine on MSE loss; a validation on `gaussian_mle` show a nearly 2X slowdown:

- Current/main branch: https://github.com/Evovest/EvoTrees.jl/blob/main-bench/benchmarks/results/regressor-cpu-gaussian.csv
- This PR: https://github.com/Evovest/EvoTrees.jl/blob/jdb/harm-v2/benchmarks/results/regressor-cpu-gaussian.csv

This may be due to some specialization used for K=1, but that don't benefit multi-target or K=2 like Gaussian-MLE.
I realize that a more systematic benchmarking should be run especially in the context of these structural change that directly affect performance. 
Can you investigate this further?

jeremiedb:
@AdityaPandeyCN Just a heads-up that syncing following fix to logistic_mle resulted in introducing back the GPU specific `_metric_value`: https://github.com/Evovest/EvoTrees.jl/blob/jdb/harm-v2/ext/EvoTreesKernelAbstractionsExt/metrics.jl#L1-L31
I've rerun the benchmarks that has no meaningful impact on perf. 

Feel free to clean them out for simplicity. In any case, just pay attention to ensure not override the fix that was done on logistic-mle metric from https://github.com/Evovest/EvoTrees.jl/pull/348

Issue #330, opened 2026-07-14
A function for leaf index extraction


EMCMS:
Hi, 

I am trying to use EvoTrees for a self-supervised task with masking where I can use the leaf index as an embedding of my data. Would it be possible to add such a function? 

Thank you in advance.

jeremiedb:
Hi, could you share some further details about your setup? It's not obvious for me at first glance for me to grasp how returning the leaf index of each tree would be relevant in a self-supervised learning context. 

If I take the request at face value, you'd like to have something like:
```julia
leaf_idx = predict_leaf_idx(model, deval)
```
where `size(leaf_idx) = (nobs, ntrees)`? Ie: for each observation, we attach the `idx` of the leaf?
I'm not too clear such leaf-idx could carry meaningful information. I'm likely confused about the nature of the task you're performing! 
In a self-learning context, I'd think of an auto-encoder setting whre the target is the input features themselves (multi-target objective), which sounds like a strange architecture for a classical greedy-tree setup; as opposed to the deep learning more typically used in such context AFAIK. 
(perhaps NeuroTabModels.jl may be oif interest here).

EMCMS:
Here is our current setup: 

cat(X, Masking) -> CatBoost --> X. 

The masking matrix is a binary matrix that tells the model what it needs to predict. We use calc_leaf_index() (https://catboost.ai/docs/en/concepts/python-reference_catboost_calc_leaf_indexes) function at the moment for this. 

This setup works very well for tabular and aligned data- much better than NN.

AJ0070:
The pattern @EMCMS describes is fairly well established outside CatBoost: scikit-learn exposes the same thing as `.apply()` on its tree ensembles, and using GBDT leaf indices as sparse categorical features for a downstream linear model goes back to He et al. 2014, *Practical Lessons from Predicting Clicks on Ads at Facebook*. The leaf id carries information as a learned bucketing of the feature space, not as a value.

Implementation looks small. The descent loop in `src/predict.jl` already computes `nid`, and it only touches `tree.split`, `tree.feat`, `tree.cond_bin` and `feattypes`, so it is loss independent: one generic method would cover all types rather than the five `predict!` variants.

@jeremiedb happy to open a PR for `predict_leaf_idx(model, data)` returning `(nobs, ntrees)` as you sketched, if you would take it.

jeremiedb:
[...]

PR #343, opened 2026-08-10
Add predict_leaf_idx for leaf index extraction

AJ0070:
Adds `EvoTrees.predict_leaf_idx(m, data; ntree_limit)`, returning a `Matrix{UInt32}` of size `(nobs, ntree_limit)` where `[i, j]` is the index of the leaf reached by observation `i` in tree `j`. Node indices follow the existing numbering: root is `1`, children of `n` are `2n` and `2n + 1`.
PR to address #330

Reuses the descent already in `predict!`, but records `nid` instead of accumulating `tree.pred`. The traversal only reads `tree.split`, `tree.feat`, `tree.cond_bin` and `feattypes`, so it is loss independent and a single generic method covers all model types rather than one per loss. No existing code paths are modified, and `predict_leaf_idx` is left unexported to match `predict` and `fit`.

Tests in `test/predict-leaf-idx.jl` cover shape and type, that every returned index is a leaf and in bounds, agreement with an independent traversal against the tree fields, consistency with `predict` (observations sharing a leaf in every tree get identical predictions), `ntree_limit` truncation and its error path, a `max_depth=1` stump, and table input matching the equivalent matrix. Docstring registered under Public API > Predict in `docs/src/api.md`.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -10,6 +10,7 @@ EvoTrees.fit
 
 ```@docs
 EvoTrees.predict
+EvoTrees.predict_leaf_idx
 ```
 
 ## SHAP
--- a/src/predict.jl
+++ b/src/predict.jl
@@ -122,6 +122,55 @@ function _predict(
     return pred
 end
 
+function predict_leaf_idx!(
+    leaf_idx::AbstractVector,
+    tree::Tree{L,K},
+    x_bin::Matrix{UInt8},
+    feattypes::Vector{Bool},
+) where {L,K}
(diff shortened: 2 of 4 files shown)


AJ0070:
@jeremiedb I already had written the code and was waiting for your approval on this!

AJ0070:
CI has not started on this one. I believe the workflows need maintainer approval for a first-time contributor.

In the meantime, locally the full suite is green at 460/460 with the added `test/predict-leaf-idx.jl`.

jeremiedb:
Good to go, thanks @AJ0070

PR #346, opened 2026-08-13
Fix gini metric: relax gini_raw signature and correct argument order

AJ0070:
`metric=:gini` throws on every call:

```
MethodError: no method matching gini_raw(::Vector{Float64}, ::SubArray{Float64, 1, Matrix{Float64}, ...})
```

There are two separate issues. `gini_raw` constrains both of its arguments to the same concrete type, but `gini` passes a view over the prediction matrix alongside a plain vector, so the call never resolves. Separately, `gini_norm` calls `gini_raw(y, p)` while `gini_raw` sorts by its first argument, so the ranking is taken from the target rather than from the prediction.

Both are fixed by matching the implementation in Evovest/MLBenchmarks.jl `src/Metrics.jl`, which has `gini_raw(p::AbstractVector, y::AbstractVector)` and `gini_raw(p, y) / gini_raw(y, y)`. The signature is also the one used here before f9210c7.

Adds `test/metrics.jl` covering the view/vector call, invariance of the normalized gini to monotone transforms of the prediction, the reversed-ranking and uninformative-prediction cases, argument order, and `metric=:gini` running end to end through `fit`. On current `main` that file produces two errors and two failures, one of the errors being `fit(...; metric=:gini)` itself.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -136,7 +136,7 @@ function multiquantile(
 end
 
 
-function gini_raw(p::V, y::V) where {V<:AbstractVector}
+function gini_raw(p::AbstractVector, y::AbstractVector)
     _y = y .- minimum(y)
     if length(_y) < 2
         return 0.0
@@ -152,7 +152,7 @@ function gini_norm(p::AbstractVector, y::AbstractVector)
     if length(y) < 2
         return 0.0
     end
-    return gini_raw(y, p) / gini_raw(y, y)
+    return gini_raw(p, y) / gini_raw(y, y)
 end
 
 function gini(
--- a/test/metrics.jl
+++ b/test/metrics.jl
@@ -1,0 +1,68 @@
+using Test
+using Statistics
+using Random
+using EvoTrees
+using EvoTrees: fit, predict, gini_raw, gini_norm
+
+@testset "metrics" begin
+
+    @testset "gini_norm" begin
+
+        nobs = 1_000
+        Random.seed!(123)
+        y = rand(nobs)
+
+        # `gini` passes a view over the prediction matrix, which is not the same concrete
+        # type as `y`. This must not throw.
+        p = reshape(copy(y), 1, :)
+        w = ones(nobs)
+        @test EvoTrees.gini(p, y, w, Float64[]) ≈ 1.0
+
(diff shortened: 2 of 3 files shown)


AJ0070:
@jeremiedb The one red job looks unrelated to this change. The suite reports `472 passed, 0 failed, 1 errored`, and the error is in `test/MLJ.jl:404`, which this PR does not touch. The other five jobs are green.

That testset is flaky by construction:

```julia
y = coerce(rand("ab", 10), OrderedFactor)
```

`rand("ab", 10)` yields a single distinct class with probability `2 * (1/2)^10`, roughly 1 run in 512. When it does, `EvoTreeClassifier` fits with `K = 1` and `predict` returns a length-`nobs` vector rather than an `nobs x K` matrix, so the `UnivariateFinite` call at `src/MLJ.jl:55` raises the `DimensionMismatch` seen in the log.

Reproduced locally:

```
two-class target -> K = 2   pred size = (10, 2)
one-class target -> K = 1   pred size = (10,)
P(all 10 draws identical) measured = 0.001935   theoretical = 0.001953
```

A rerun should be green. Happy to open a separate PR seeding that testset so it stops firing at random, and to look at the `K = 1` path as well if a single-class target is something you want supported rather than rejected.

AJ0070:
One follow-up: the `codecov/project` red is downstream of the same flaky job, not a real coverage change.

The workflow runs the coverage steps unconditionally after the test step:

```yaml
- uses: julia-actions/julia-runtest@v1
- uses: julia-actions/julia-processcoverage@v1
- uses: codecov/codecov-action@v7
```

so when the test step fails the two after it are skipped. On the failing job:

```
failure  Run julia-actions/julia-runtest@v1
skipped  Run julia-actions/julia-processcoverage@v1
skipped  Run codecov/codecov-action@v7
```

No lcov was uploaded from the Linux runner, so the project total came out at 45.55% (-6.36%). `codecov/patch` passed, which is the check that actually looks at the lines added here.

A re-run of that job should clear both. Adding `if: always()` to the two coverage steps would also stop a single flaky test from taking the coverage report with it, though that is a separate change and not something I have touched here.

jeremiedb:
Looks good, thanks for the fix. I'll push the fix to the flaky test directly following merge.

jeremiedb (approved):

Issue #345, opened 2026-08-11
MLJ update continues boosting when hyperparameters other than nrounds change


AJ0070:
`MMI.update` continues boosting an existing model whenever `nrounds` has not decreased, regardless of what else changed. `okay_to_continue` compares only `nrounds`:

```julia
function okay_to_continue(model, fitresult, cache)
  check = model.nrounds - fitresult.info[:nrounds] >= 0
  return check
end
```

So if any other hyperparameter changes while `nrounds` stays equal, the check passes, the `while` loop has nothing left to do, and the previously fitted model is returned untouched. Since MLJ calls `update` rather than `fit` whenever it can, a `TunedModel` grid over `eta` or `max_depth` ends up scoring the same fitted model repeatedly.

Changing `eta` from 0.05 to 0.9 and `max_depth` from 3 to 6, with `nrounds` held at 10:

```
rmse(update with new params) = 0.371416
rmse(fresh fit, same params) = 0.064510
update output identical to the old model?  true
update output identical to a fresh fit?    false
```

```julia
import MLJModelInterface as MMI
fr, cache, _ = MMI.fit(EvoTreeRegressor(nrounds=10, eta=0.05, max_depth=3), 0, X, y)
m2 = EvoTreeRegressor(nrounds=10, eta=0.9, max_depth=6)
fr2, _, _ = MMI.update(m2, 0, fr, cache, X, y)   # returns the old model unchanged
```

The block added in response to #92 at `test/MLJ.jl:286-295` already exercises this exact path, changing `model.lambda` before calling `update`, but makes no assertion about the result, so it passes either way.

I have a fix ready if you would like it. It keeps everything inside `src/MLJ.jl`: `MMI.fit` returns `(core=cache, model=deepcopy(model))` so `update` can compare against the model that was actually fitted, and `okay_to_continue` becomes `MMI.is_same_except(model, cache.model, :nrounds) && model.nrounds >= fitresult.info[:nrounds]`. That avoids adding a field to the Cache structs, which #340 is currently reworking.

Continuing from an existing ensemble is preserved for the `nrounds`-only case, and I checked that growing 10 to 25 rounds through `update` gives results identical to a fresh 25-round fit. The added tests fail on current `main` with two failures and pass with the fix.

Happy to open the PR, or to take a different approach if you would rather the previous model live somewhere else.

ablaom:
[...]

Issue #344, opened 2026-08-10
Unexpected behaviour in learning curves and tuning of hyperparameters 


ablaom:
Is this expected?

```
using MLJ
EvoTreeRegressor = @load EvoTreeRegressor
model = EvoTreeRegressor()
X, y = make_regression()
curve = learning_curve(model, X, y; range=range(model, :max_depth, lower=1, upper=10))      
curve.measurements
# 10-element Vector{Float64}:
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
# 0.390402111939076
```

```
Status `~/GoogleDrive/Julia/MLJ/MLJTutorial/docs/src/notebooks/99_solution_to_exercises/Project.toml`
  [336ed68f] CSV v0.10.16
  [a93c6f00] DataFrames v1.8.2
  [31c24e10] Distributions v0.25.130
  [f6006082] EvoTrees v0.18.7
  [add582a8] MLJ v0.23.3
  [d354fa79] MLJClusteringInterface v0.1.13
  [c6f25543] MLJDecisionTreeInterface v0.5.0
  [6ee0df7b] MLJLinearModels v0.10.2
  [91a5bcdd] Plots v1.41.6
  [2913bbd2] StatsBase v0.34.12
  [f43a241f] Downloads v1.7.0
```

ablaom:
On further investigation, perhaps the issue is on the MLJ side. Closing for now.

ablaom:
I really don't know yet where this is coming from.  I am only getting this for EvoTreeRegressor - not other tree models (e.g DecisionTreeRegressor from DecisionTree.jl). The mystery is that I can mutate `max_depth` by hand and I do see an effect:

```julia
julia> evaluate(model, X, y).measurement[1]
0.17637116508907527

julia> model.max_depth = 2
2

julia> evaluate(model, X, y).measurement[1]
0.19007430389649682
```

jeremiedb:
I'm unable to perform test immediatly, but I can confirm the parameters mutation is effective.
Could this be related to the copy of parameters returned; for example the issue initially experiences where the original seed was not used and instead the current training state was kept?

ablaom:
[...]

Issue #92, opened 2021-04-15
Bug in `MLJModelInterface.update`


ablaom:
Ran into a bug when trying to tune a EvoTreeRegressor in MLJ. Isolated it to this:

```julia
using MLJ
using EvoTrees
using MLJModelInterface
const MMI = MLJModelInterface

X, y = @load_boston
model = (@load EvoTreeRegressor)()
data = MMI.reformat(model, X, y)
f, c, r = MMI.fit(model, 2, data...);
model.λ = 0.1

julia> MMI.update(model, 2, f, c, data...);
ERROR: ArgumentError: Function `matrix` only supports AbstractMatrix or containers implementing the Tables interface.
Stacktrace:
 [1] matrix(::FullInterface, ::Val{:other}, X::NamedTuple{(:matrix, :names), Tuple{Matrix{Float64}, Vector{Symbol}}}; kw::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ MLJModelInterface ~/.julia/packages/MLJModelInterface/tegnW/src/data_utils.jl:32
 [2] matrix(::FullInterface, ::Val{:other}, X::NamedTuple{(:matrix, :names), Tuple{Matrix{Float64}, Vector{Symbol}}})
   @ MLJModelInterface ~/.julia/packages/MLJModelInterface/tegnW/src/data_utils.jl:32
 [3] matrix(X::NamedTuple{(:matrix, :names), Tuple{Matrix{Float64}, Vector{Symbol}}}; kw::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ MLJModelInterface ~/.julia/packages/MLJModelInterface/tegnW/src/data_utils.jl:27
 [4] matrix
   @ ~/.julia/packages/MLJModelInterface/tegnW/src/data_utils.jl:27 [inlined]
 [5] reformat(#unused#::EvoTrees.EvoTreeRegressor{Float64, EvoTrees.Linear, Int64}, X::NamedTuple{(:matrix, :names), Tuple{Matrix{Float64}, Vector{Symbol}}}, y::SubArray{Float64, 1, Matrix{Float64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, true})
   @ EvoTrees ~/.julia/packages/EvoTrees/oHLKA/src/MLJ.jl:24
[...]

PR #347, opened 2026-08-14
Fix MLJ update to refit when hyperparameters other than nrounds change

AJ0070:
`MMI.update` continues boosting an existing model whenever `nrounds` has not decreased, regardless of what else changed, because `okay_to_continue` compares only `nrounds`. Any other hyperparameter change returns the previously fitted model untouched, so MLJ tuning silently scores the same model repeatedly.

PR to address #345, and #344 which has the same cause.

`MMI.fit` now returns the fitted model alongside the cache, so `update` can compare against what was actually fitted:

```julia
return fitresult, (core=core, model=deepcopy(model)), report

function okay_to_continue(model, fitresult, cache)
  return MMI.is_same_except(model, cache.model, :nrounds) &&
         model.nrounds >= fitresult.info[:nrounds]
end
```

Keeping the previous model in the MLJ cache rather than on the Cache structs confines the change to `src/MLJ.jl`, which avoids overlapping #340. It follows the same shape as MLJDecisionTreeInterface, including the `verbosity > 0 && @info` on cold restart.

Continuing from an existing ensemble is preserved for the `nrounds`-only case: growing 10 rounds to 25 through `update` gives predictions identical to a fresh 25-round fit.

One deliberate difference from the MLJDecisionTreeInterface example: it handles a decrease in the iteration parameter by truncating the existing forest. EvoTrees pushes `bagging_size` trees per round, so the equivalent would need `1:(nrounds * bagging_size)` plus a matching `info[:nrounds]` update. That is an optimisation rather than a correctness fix, so a decrease still goes through a full refit here.

Verified against the `learning_curve` reproduction in #344, sweeping `max_depth` from 1 to 10:

```
main:  [3.448369, 3.448369, 3.448369, ...]   1 of 10 distinct
fix:   [3.448369, 0.396577, 0.314992, ...]  10 of 10 distinct
```

Tests added in `test/MLJ.jl` cover a hyperparameter change refitting and matching a fresh fit, a decrease in `nrounds` refitting, and an increase continuing while still agreeing with a fresh fit. They produce two failures on current `main`. They sit alongside the block added in response to #92, which already exercised this path but asserted nothing about the result.

### DIFF
--- a/src/MLJ.jl
+++ b/src/MLJ.jl
@@ -8,18 +8,28 @@ function MMI.fit(model::EvoTypes, verbosity::Int, A, y, w=nothing)
   device = model.device == :gpu ? GPU : CPU
   V = device_array_type(device)
   w = isnothing(w) ? device_ones(device, T, nobs) : V{T}(w)
-  fitresult, cache = init_core(model, device, A, feature_names, y, w, nothing)
+  fitresult, core = init_core(model, device, A, feature_names, y, w, nothing)
 
   while fitresult.info[:nrounds] < model.nrounds
-    grow_evotree!(fitresult, cache, model)
+    grow_evotree!(fitresult, core, model)
   end
-  report = (features=cache.feature_names,)
-  return fitresult, cache, report
+  report = (features=core.feature_names,)
+  return fitresult, (core=core, model=deepcopy(model)), report
 end
 
+"""
+    okay_to_continue(model, fitresult, cache)
+
+Whether `MMI.update` may extend the existing fit instead of retraining from scratch.
+
+Boosting can only append trees to an existing ensemble, so continuing is valid only when the
+requested model differs from the fitted one in `nrounds` alone, and `nrounds` has not been
+reduced below the number of trees already grown. Any other hyperparameter change requires a
+cold restart, since the trees already fitted were grown under the previous values.
+"""
 function okay_to_continue(model, fitresult, cache)
-  check = model.nrounds - fitresult.info[:nrounds] >= 0
-  return check
+  return MMI.is_same_except(model, cache.model, :nrounds) &&
+         model.nrounds >= fitresult.info[:nrounds]
 end
 
 # For EarlyStopping.jl support
@@ -36,11 +46,12 @@ function MMI.update(
 )
   if okay_to_continue(model, fitresult, cache)
     while fitresult.info[:nrounds] < model.nrounds
-      grow_evotree!(fitresult, cache, model)
+      grow_evotree!(fitresult, cache.core, model)
     end
-    report = (features=cache.feature_names,)
+    cache = (core=cache.core, model=deepcopy(model))
+    report = (features=cache.core.feature_names,)
   else
-    fitresult, cache, report = fit(model, verbosity, A, y, w)
+    fitresult, cache, report = MMI.fit(model, verbosity, A, y, w)
   end
   return fitresult, cache, report
 end
--- a/src/MLJ.jl
+++ b/src/MLJ.jl
@@ -51,6 +51,8 @@ function MMI.update(
     cache = (core=cache.core, model=deepcopy(model))
     report = (features=cache.core.feature_names,)
   else
+    verbosity > 0 &&
+      @info "Only `nrounds` supports warm restart. Retraining from scratch."
     fitresult, cache, report = MMI.fit(model, verbosity, A, y, w)
   end
   return fitresult, cache, report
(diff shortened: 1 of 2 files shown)


AJ0070:
The macOS red on the previous run was the `MLJ - supported ordered factor predictions` testset, and it was my fault rather than chance.

`test/core.jl:8` seeds the global RNG, so the whole suite runs on a deterministic stream. My testset was drawing from that global RNG for its fixtures, which consumed 800 values and shifted the stream for every test that follows. That testset builds its target with

```julia
y = coerce(rand("ab", 10), OrderedFactor)
```

which yields a single distinct class roughly 1 run in 512. On the shifted stream, macOS landed on exactly that: `EvoTreeClassifier` then fits with `K = 1` and `predict` returns a length-`nobs` vector where `UnivariateFinite` expects `nobs x K`, giving the `DimensionMismatch` in the log. The suite still reported `477 passed, 0 failed, 1 errored`, so every assertion in this PR passed.

Fixed by giving the testset its own RNG, so it consumes nothing global. Verified that the downstream stream is now byte-identical to `main`:

```
drawing from global : abbabaaabb
local rng           : bbbabaabba
main baseline       : bbbabaabba
```

Worth noting independently of this PR: because that testset draws an unseeded target from the shared stream, any new test inserted before it can flip it red without being related to it. Seeding it, or constructing the target so both classes are guaranteed present, would remove the trap. Happy to send that as a separate PR if useful.

ablaom:
Thanks @AJ0070 for the substantial work here. I can take a look early next week.

jeremiedb (approved):

jeremiedb:
Thanks @AJ0070, looks good to me. I'll for feedbacks @ablaom may have, and otherwise will proceed with merge. 

Regarding flaky test, I've replaced them with a deterministic input: https://github.com/Evovest/EvoTrees.jl/blob/ae1762a2db03dfffcf697126f1dae09648670697/test/MLJ.jl#L406. It's already on main.

src/MLJ.jl
```diff
@@ -36,11 +46,14 @@ function MMI.update(
...
-      grow_evotree!(fitresult, cache, model)
+      grow_evotree!(fitresult, cache.core, model)
     end
-    report = (features=cache.feature_names,)
+    cache = (core=cache.core, model=deepcopy(model))
+    report = (features=cache.core.feature_names,)
   else
-    fitresult, cache, report = fit(model, verbosity, A, y, w)
+    verbosity > 0 &&
+      @info "Only `nrounds` supports warm restart. Retraining from scratch."
```
ablaom: I suggest dropping this bit of logging, as there is already [generic logging](https://github.com/JuliaAI/MLJBase.jl/blob/6b9ef82b4069298a6fcb691882eb2438f4907176/src/machines.jl#L538) at a the higher level of `fit!(::Machine, ...)` in MLJBase.

ablaom (commented):
Just the one suggestion re logging. Otherwise looks good. 

Very good to get your diagnosis @AJ0070 - saved me some work. It's wild that this bug is around 5 years old and we both tripped over it within a few days of each other.

src/MLJ.jl
AJ0070: Done, thanks. Dropped in 41ae266.

Confirmed the generic path covers it: `fitlog` logs `Training $mach` when the machine retrains, so a cold restart is already visible without EvoTrees saying anything of its own.

One observation while checking, not a reason to keep the line. Because the fallback happens inside `MMI.update`, MLJBase logs `Updating $mach` even though a full refit just occurred, so the message a user sees is the opposite of what happened. Nothing this PR should change, and only relevant when a non-iteration hyperparameter has moved, but it seemed worth mentioning given you maintain that side.

ablaom: > One observation while checking ...

Good point. Flagged [here](https://github.com/JuliaAI/MLJBase.jl/issues/1065)

ablaom (approved):
All good.

ablaom:
@jeremiedb I think this is ready to roll.

AJ0070:
@jeremiedb

ablaom:
@jeremiedb Any chance of a patch release soon to get this bug fix? In particular, this is holding back some documentation fixes/updates.

jeremiedb:
@ablaom v0.19 has just been released!
https://github.com/JuliaRegistries/General/pull/165920

ablaom:
Thanks @jeremiedb.

A minor note: I would not have characterized  this PR as "breaking". It is a bug fix.

PR #348, opened 2026-08-14
Fix logistic_mle eval metric to match the logistic log-density

AJ0070:
The `logistic_mle` eval metric does not match the objective it reports on. The logistic density with location `mu` and scale `s` is `(1/4s) * sech((y-mu)/(2s))^2`, but `_mle2p_metric_value` evaluates `sech((y-mu)/s)`, dropping the factor of 2 in the argument. The gradient in `mle2p_grad_hess` uses the correct `(yt - μ) / (2 * exp(ls))`, so the reported metric and the optimised loss disagree.

With `mu=0.5` and `s=2.0` the metric returns -2.319672 at `y=1.5` and -4.252654 at `y=4.0`, against a true log-density of -2.141301 and -2.763595. They agree only at `y = mu`, where the argument is zero, and diverge as the residual grows. Since `is_maximise(logistic_mle) = true` drives early stopping, training also stops on the wrong signal.

Fixed in both the CPU and CUDA copies. After the fix the metric matches the closed form exactly, and its derivative agrees with the gradient the loss uses: `d(metric)/dmu = 0.122459` against `-g1 = 0.122459`, where the previous value was -0.462117.

Adds tests in `test/metrics.jl` checking the metric against the closed-form density over several location and scale combinations, that its numerical derivative matches `mle2p_grad_hess`, and that it is symmetric about the location and maximised there. They fail on current `main`.

### DIFF
--- a/ext/EvoTreesCUDAExt/metrics.jl
+++ b/ext/EvoTreesCUDAExt/metrics.jl
@@ -28,7 +28,7 @@ end
 end
 
 @inline _mle2p_metric_value(::Type{EvoTrees.GaussianMLE}, μ, ls, yt) = -(ls + (yt - μ)^2 / (2 * exp(2 * ls)))
-@inline _mle2p_metric_value(::Type{EvoTrees.LogisticMLE}, μ, ls, yt) = log(1 / 4 * sech(exp(-ls) * (yt - μ))^2) - ls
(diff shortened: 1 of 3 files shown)


jeremiedb (approved):

jeremiedb:
Thanks for the catch! Good idea to include validation check on the derivative.

PR #349, opened 2026-08-14
Error on a single-level classification target

AJ0070:
A single-level classification target currently produces a degenerate model rather than an error. `K` ends up as 1, and predicting through the MLJ interface then fails with `DimensionMismatch: Probability array is incompatible with the number of classes` out of `UnivariateFinite`, which gives no indication of the actual problem.

Following your preference for an upstream control, `init_core` now rejects it at fit:

```
Classification requires a target with at least 2 levels, got 1: ["a"]. A single-class problem is not meaningful.
```

This is the underlying cause of the ordered factor test flakiness patched in dc9aabd, where `rand("ab", 10)` occasionally produced a target with a single class.

Adds tests for the error on both a plain vector and a categorical target, that two and three level targets are unaffected, and the MLJ path with a single-level target.

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -55,6 +55,10 @@ function init_core(params::EvoTypes, ::Type{CPU}, data, feature_names, y_train, 
             @error "Invalid target eltype: $(eltype(y_train))"
         end
         K = length(target_levels)
+        K < 2 && error(
+            "Classification requires a target with at least 2 levels, got $K: " *
+            "$(string.(target_levels)). A single-class problem is not meaningful."
+        )
         μ = T.(log.(proportions(y, UInt32(1):UInt32(K))))
         μ .-= maximum(μ)
         !isnothing(offset) && (offset .= log.(offset))
--- a/test/MLJ.jl
+++ b/test/MLJ.jl
@@ -409,3 +409,10 @@ end
     yhat = predict(mach, X)
     @assert isordered(yhat)
 end
+
+@testset "MLJ - single-class target" begin
+    X = (; x=rand(Xoshiro(11), 10))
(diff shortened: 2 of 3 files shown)


AJ0070:
@jeremiedb worth noting this makes the ordered factor testset robust regardless of how its target is drawn. With `_predict` returning `nobs x K` for `MLogLoss`, the earlier `rand("ab", 10)` version would pass even on a draw that produced a single class, so the flakiness is fixed at the source rather than only in the fixture.

Same applies to anything else that can land on a degenerate target, a CV fold or a filtered subset for instance, which currently fails with a `DimensionMismatch` from inside `UnivariateFinite` rather than anything readable.

jeremiedb:
For multi-classification; I'd prefer to have an upstream control and error if the target has a single level. I think it's a legitimate exception as a single class problem is not meaningful.

jeremiedb (approved):

AJ0070:
@jeremiedb one thing I noticed while looking around: `EvoTrees.save` and `EvoTrees.load` have no assertions anywhere in the suite. `test/save_load.jl` exists but is not included in `runtests.jl` and contains no `@test`, and the same is true of `weighted.jl`, `issue-243.jl` and `issue-274.jl`, so none of them run in CI.

To be clear this is not a bug report: I checked and the round trip is correct today across regressor, logloss, count, gaussian MLE and classifier, with predictions identical after a save and load cycle.

Happy to add a round-trip test over those model types, asserting predictions and the `info` fields survive, given BSON tends to be sensitive to version changes. Equally happy to leave it if you would rather not carry the extra test time.

PR #351, opened 2026-08-16
Return zero importance when a model contains no splits

AJ0070:
`EvoTrees.importance` returns `NaN` for every feature when a model contains no split. The gain vector is normalised with `gain .= gain ./ sum(gain)`, and that total is zero in such a model.

Three realistic paths reach it: `max_depth=1`, `nrounds=0`, and a `gamma` high enough that no candidate split is accepted.

```julia
m = fit(EvoTreeRegressor(nrounds=5, max_depth=1); x_train=x, y_train=y)
EvoTrees.importance(m)
# [:feat_1 => NaN, :feat_3 => NaN, :feat_4 => NaN, :feat_2 => NaN]
```

It reaches MLJ too, where `feature_importances` returns `[:a => NaN, :b => NaN, :c => NaN]`.

The zeros are now returned as they are when the total gain is zero.

This also replaces `collect(Dict(zip(...)))` with direct pair construction. The dictionary round trip returned ties in hash order rather than feature order, so repeated calls on the same model could disagree, and duplicate names passed through `feature_names` were silently collapsed.

Tests cover the three no-split paths, that a model which does split still normalises to one, ordering stability across calls, and that duplicate names are preserved. They fail on current `main`.

### DIFF
--- a/src/importance.jl
+++ b/src/importance.jl
@@ -19,8 +19,12 @@ function importance(model::EvoTree; feature_names=model.info[:feature_names])
         importance!(gain, tree)
     end
 
-    gain .= gain ./ sum(gain)
-    pairs = collect(Dict(zip(Symbol.(feature_names), gain)))
+    # A model can contain no split at all (`max_depth = 1`, `nrounds = 0`, or a `gamma` high
+    # enough to reject every candidate), in which case the total gain is zero and normalising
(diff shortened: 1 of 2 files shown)


jeremiedb:
Thanks for this robustness improvement!

jeremiedb (approved):

PR #355, opened 2026-08-17
Apply the single-level target check to the GPU init path

singhharsh1708:
PR #349 rejects a classification target with fewer than 2 levels, but the check only landed in `src/init.jl`. `EvoTrees.init_core` for `GPU` in `ext/EvoTreesCUDAExt/init.jl` is a separate method and still runs the old path, so `device=:gpu` builds the degenerate single-class model that #349 was meant to prevent.

```
$ git show --stat --format="" feed3ea
 src/init.jl  |  4 ++++
 test/MLJ.jl  |  7 +++++++
 test/core.jl | 23 ++++++++++++++++++++++-

$ grep -c "at least 2 levels" src/init.jl ext/EvoTreesCUDAExt/init.jl
ext/EvoTreesCUDAExt/init.jl:0
src/init.jl:1
```

This copies the check to the same position in the GPU branch, right after `K = length(target_levels)`, with the same message, so both devices fail the same way.

I have no GPU here so I could not exercise the GPU path itself. What I did check is that the extension still compiles with the change, by loading EvoTrees with CUDA, KernelAbstractions and Atomix present so the extension is triggered. The CPU tests added in #349 already cover the behaviour and CI has no GPU runner, so I did not add a test.

### DIFF
--- a/ext/EvoTreesCUDAExt/init.jl
+++ b/ext/EvoTreesCUDAExt/init.jl
@@ -53,6 +53,10 @@ function EvoTrees.init_core(params::EvoTrees.EvoTypes, ::Type{<:EvoTrees.GPU}, d
             @error "Invalid target eltype: $(eltype(y_train))"
         end
         K = length(target_levels)
+        K < 2 && error(
+            "Classification requires a target with at least 2 levels, got $K: " *
+            "$(string.(target_levels)). A single-class problem is not meaningful."
+        )
         μ = T.(log.(EvoTrees.proportions(y, UInt32(1):UInt32(K))))
         μ .-= maximum(μ)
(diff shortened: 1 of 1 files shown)


jeremiedb:
Could you check conflicts with main?

singhharsh1708:
Checked. The conflict is that `ext/EvoTreesCUDAExt/init.jl` was deleted by the KernelAbstractions migration, so there is nothing left for this patch to apply to.

The GPU path now calls `EvoTrees._init_target` (`src/init.jl:11` from `ext/EvoTreesKernelAbstractionsExt/init.jl`), and that function already carries the `K < 2` check at `src/init.jl:56`. Both devices go through the same guard now, so the gap this was fixing is closed by the refactor. Closing it.

Issue #356, opened 2026-08-19
Early stopping locates the best iteration but the returned model ignores it


AJ0070:
When early stopping fires, training halts but every tree grown past the best iteration stays in the model, and `predict` uses all of them by default. `logger[:best_iter]` is recorded and then never applied.

The counterintuitive part is that the damage grows with `early_stopping_rounds`. Holding everything else fixed and varying only the patience:

```
patience  trees  best_iter  rmse_full  rmse@best     gap
10        17     6          1.52684    1.52136       0.00548
25        32     6          1.53666    1.52136       0.01530
50        57     6          1.55011    1.52136       0.02875
100       107    6          1.58328    1.52136       0.06192
```

`best_iter` is stable at 6 and `rmse@best` is identical every time, so the optimum is always located correctly. Only the returned model degrades, and roughly linearly in the patience. Raising `early_stopping_rounds` to be more confident of finding the optimum therefore makes the delivered model worse, which inverts what the parameter is for.

Recovering the best model is possible today but requires knowing to do it:

```julia
bagging = (length(m.trees) - 1) ÷ m.info[:nrounds]
p = EvoTrees.predict(m, x; ntree_limit = 1 + m.info[:logger][:best_iter] * bagging)
```

There is a design choice here and I do not want to presume:

1. Truncate `m.trees` at the best iteration at the end of `fit`, and set `info[:nrounds]` to match. The model then is what early stopping selected, and it is smaller. Destructive, in that the post-optimum trees are discarded.
2. Keep every tree and have `predict` default to the best iteration. Non-destructive, but `length(m.trees)` no longer reflects what `predict` uses.

XGBoost takes the second route, exposing `best_iteration` alongside `iteration_range`.

I have a PR ready for option 1 and am happy to switch it to option 2 if you prefer, or to leave it if you would rather the current behaviour stand and be documented instead.

PR #357, opened 2026-08-19
Truncate the model to the best iteration when early stopping fires

AJ0070:
Trees grown after the best iteration are dropped when early stopping fires, so the returned model is the one early stopping selected rather than the one at the stopping point.

PR to address #356. This implements option 1 from that issue. I can switch to option 2 instead, keeping every tree and having `predict` default to the best iteration.

Holding everything fixed and varying only `early_stopping_rounds`, test rmse before and after:

```
patience   before     after
10         1.52684    1.52136
25         1.53666    1.52136
50         1.55011    1.52136
100        1.58328    1.52136
```

`best_iter` was already stable at 6 in every one of those runs, so the optimum was always located correctly; only the delivered model varied.

`truncate_to_best_iter!` is a no-op when no eval set was supplied, when the best iteration is the last one, and when early stopping never fired. That last case matters on its own: an eval set is often passed for monitoring with no intention of stopping early, and such a fit has to return every tree that was asked for. The check is `iter_since_best >= early_stopping_rounds`.

It derives `bagging_size` from `length(m.trees) ÷ nrounds`, so the mapping from rounds to trees holds when more than one tree is grown per round, and it sets `info[:nrounds]` to match the retained trees.

Tests cover the four patience values agreeing, the tree count and `nrounds` matching `best_iter`, a model fitted without an eval set keeping every tree, an eval set with early stopping never firing keeping every tree and its last logged metric still matching the returned model, `bagging_size` at 1 and 3, and a run that improves to the final round.

Rebased on current main. #384 moved the intercept off `trees[1]`, so `trees` now holds `nrounds * bagging_size` entries and the round-to-tree arithmetic here drops the offset it used to carry.

### DIFF
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -339,9 +339,38 @@ function fit(
     end
     post_fit_gc(_device)
     m.info[:logger] = logger
+    truncate_to_best_iter!(m)
 
     return m
 
+end
+
+
+"""
+    truncate_to_best_iter!(m::EvoTree)
+
+Drop the trees grown after the best iteration recorded by early stopping.
+
+Early stopping keeps boosting for `early_stopping_rounds` after the optimum in order to confirm
+it, and those extra rounds are overfit by construction. Without this the model returned is the
+one at the stopping point rather than the one early stopping selected, and the gap grows with
+`early_stopping_rounds`.
+
+No-op when no evaluation set was supplied, when early stopping never fired, or when the best
+iteration is the last one.
+"""
+function truncate_to_best_iter!(m::EvoTree)
+    logger = get(m.info, :logger, nothing)
+    isnothing(logger) && return m
+    nrounds = m.info[:nrounds]
+    best_iter = logger[:best_iter]
+    (best_iter <= 0 || best_iter >= nrounds) && return m
+    # the first tree carries the bias, the remainder are `bagging_size` per round
+    bagging_size = (length(m.trees) - 1) ÷ nrounds
+    keep = 1 + best_iter * bagging_size
+    resize!(m.trees, keep)
+    m.info[:nrounds] = best_iter
+    return m
 end
 
 """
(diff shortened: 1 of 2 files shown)


jeremiedb:
I have mixed feelings about the ideal route to take. A consideration I'd have is that the current approach is consistent with the behavior associated with early stopping on a neural-net (ex: NeuroTabModels). Since there's not a systematic checkpointing of parameters; the resulting model is in the state as it was after the waiting period. 
I agree though that returning the model in its state as-of the best iteration has appeal. A design question I have is that the definition of a best iteration may differ if one wants to consider running mean/median to avoid picking a single flaky eval point. 

I'd leave the PR at the moment; until a clear design is laid out and avoid back-and-forth design changes. 
In any case, it will be a good thing to explicitly state the contract regarding early stopping in EvoTrees.

AJ0070:
Happy to leave it parked.

One difference from the neural-net case that might narrow the design space: because boosting is additive, the best-iteration model is already an exact prefix of the returned one, so nothing needs checkpointing. On current main `predict(m, x; ntree_limit=best_iter+1)` reproduces `logger[:best_metric]` exactly:

```
rowsample=1.0   best_iter=85   logger 0.15890233   predict @ ntree_limit 0.15890233
rowsample=0.5   best_iter=74   logger 0.15875153   predict @ ntree_limit 0.15875153
```

So the best model is always recoverable from what `fit` returns, and the question is only which one `predict` gives by default, not whether the good one is reachable. That also means this is not a one-way door: keeping every tree and defaulting `ntree_limit` to `best_iter` (option 2 in #356) preserves the returned object and the NeuroTabModels parallel, while making the default match what early stopping selected.

On the running mean/median point, I'd read that as orthogonal: it changes how `best_iter` is computed, not what the model does with it. The logger already keeps the full metric history, so a smoothed rule could replace the current single-point argmin later without revisiting the truncation decision.

Agreed the contract is the thing worth pinning down first. Happy to write that up as a separate docs PR covering what `best_iter` means, which model `fit` returns, and how to get the other one, and leave this one parked behind it.

PR #361, opened 2026-08-20
Encode eval target against the model's training levels

AJ0070:
The eval callback builds its class encoding from `y_eval` itself rather than from the levels the model was trained on. `m.info[:target_levels]` is written at fit and then read nowhere except `src/MLJ.jl:55`, so an eval set whose level set differs from the training one is scored against the wrong prediction columns.

The resulting integer is used directly as a row index into the prediction matrix in `mlogloss`, so the damage is silent. Three classes, 30 rounds, comparing the reported metric against the same quantity computed by hand through `target_levels`:

```
eval contains            reported    true
all 3 classes            0.18970     0.18970
missing class 1          3.25376     0.15333
missing class 2          2.85033     0.17334
only class 3             4.45472     0.11401
```

A class-complete eval set agrees, which is why this is easy to miss. A stratified split that drops a rare class, or a fold of an imbalanced multiclass problem, does not.

Since `is_maximise(mlogloss) = false` drives early stopping, training also stops on the wrong signal. With `nrounds=200` and `early_stopping_rounds=10` against an eval fold missing one class, the logger reports `best_iter = 0` and stops after 10 rounds, so the iteration recorded as best is the bias-only model.

Separately, a level present in `y_eval` but absent from `y_train` indexes past the end of the prediction matrix. `mlogloss` reads `p[y[i], i]` under `@inbounds`, so on a two-class model an eval label of 3 reads out of bounds and reports whatever it finds:

```
default (@inbounds active):   no error, metric 1.5913
--check-bounds=yes:           BoundsError, "attempt to access 2x2000 Matrix"
```

`eval_levelcode` now maps `y_eval` through `m.info[:target_levels]` with `indexin`, and rejects a level that never appeared in training rather than indexing past the end. This works uniformly for integer, string and categorical targets, since a raw value compares equal to its `CategoricalValue` level. Both `CallBack` constructors are covered, and `CallBack` is shared across backends so this is not CPU only. The dead `levels` and `μ` locals go with it.

Adds a testset in `test/metrics.jl` checking the reported metric against a direct computation for a class-complete eval set and two class-missing ones, and that an unseen eval level throws. The class-complete case passes either way, so the missing-class cases are what separate the two encodings.

### DIFF
--- a/src/callback.jl
+++ b/src/callback.jl
@@ -1,3 +1,23 @@
+_raw_level(x) = x isa CategoricalValue ? CategoricalArrays.unwrap(x) : x
+
+"""
+    eval_levelcode(y_eval, target_levels)
+
+Encode `y_eval` against the class levels the model was trained on, rather than against the
+levels present in `y_eval` itself. An eval set that is missing a training class, or that
+orders its own levels differently, would otherwise be scored against the wrong prediction
+columns.
+"""
+function eval_levelcode(y_eval, target_levels)
+    idx = indexin(y_eval, target_levels)
+    if any(isnothing, idx)
+        unseen = unique(_raw_level.(y_eval[findall(isnothing, idx)]))
+        error("`y_eval` contains levels absent from `y_train`: $(unseen). " *
+              "Training levels are $(_raw_level.(target_levels)).")
+    end
+    return UInt32.(idx)
+end
+
 struct CallBack{B,P,Y,C,D,K}
     feval::Function
     x_bin::B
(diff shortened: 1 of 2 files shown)


jeremiedb:
Good catch! Could you resolve conflicts, then good to go!

AJ0070:
Rebased on current main, conflict was the same one as #360: #358 replaced the `logistic_mle metric` testset with `MLE metrics`, so I kept yours and placed the new testset above it. The `src/callback.jl` change is unaffected and still applies, main still derives the eval encoding from `sort(unique(y_eval))`.

jeremiedb:
Could you sync back, seems like the conflicts reappeared following merge of #360

AJ0070:
Synced. Same spot again: #360 landed its `quantile metric alpha` testset next to where this one goes, so both now sit side by side. `src/callback.jl` is unaffected and still applies, main still derives the eval encoding from `sort(unique(y_eval))`. Full suite green at 622.

PR #362, opened 2026-08-20
Test that multi-target MLE eval offsets unconstrain every scale column

AJ0070:
Rebased on current main. Worth flagging that this changed what the PR is: #358 introduced `unconstrain_mle_scale!`, which uses `offset[:, 2:2:end]` and is now called from both `CallBack` constructors, so the source fix here is already in. I have dropped it and kept only the regression test.

Verified against current main that the eval callback now unconstrains every scale column, not just the second: with a four column offset at `s1 = 1.2` and `s2 = 1.5`, `cb.p[2, 1]` and `cb.p[4, 1]` both match `invsoftplus`. Before #358 column 4 stayed raw at 1.5. The test also pins that a two column offset is unchanged, since `2:2:end` reduces to `2:2` there.

Retitled accordingly. Happy to close this instead if you would rather not carry the extra test.

### DIFF
--- a/test/multi-target.jl
+++ b/test/multi-target.jl
@@ -50,4 +50,29 @@ using EvoTrees: fit, predict, sigmoid, logit
         @test mean((pred[:, 1] .- dtrain.y).^2)  < base[1] * 0.9
         @test mean((pred[:, 3] .- dtrain.y2).^2) < base[2] * 0.9
     end
+
+    # Multi-target MLE packs [mu_1, phi_1, mu_2, phi_2, ...], so every even offset column
+    # is a scale and must be unconstrained, not just column 2. The eval callback builds
+    # its own copy of this mapping, so it is worth pinning that it covers all of them.
+    @testset "multi-target MLE eval offset" begin
+        n_off = 200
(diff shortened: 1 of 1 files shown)


jeremiedb:
Looks good. Could you sync with the current main to resolve conflict?

AJ0070:
Rebased on current main. Worth flagging that this changed what the PR is: #358 introduced `unconstrain_mle_scale!`, which uses `offset[:, 2:2:end]` and is now called from both `CallBack` constructors, so the source fix here is already in. I have dropped it and kept only the regression test.

Verified against current main that the eval callback now unconstrains every scale column, not just the second: with a four column offset at `s1 = 1.2` and `s2 = 1.5`, `cb.p[2, 1]` and `cb.p[4, 1]` both match `invsoftplus`. Before #358 column 4 stayed raw at 1.5. The test also pins that a two column offset is unchanged, since `2:2:end` reduces to `2:2` there.

Retitled accordingly. Happy to close this instead if you would rather not carry the extra test.

jeremiedb (approved):
Always worth having it tested, thanks!

PR #363, opened 2026-08-20
Throw instead of logging on invalid input

AJ0070:
`@error` logs and returns, it does not interrupt control flow, so each of the four places it is used as a rejection is a fall-through into a worse failure.

`subsample` is the clearest case. The guard is also testing the wrong variable:

```julia
counts_sum = sum(counts)
if counts_cum == 0
```

`counts_cum` is `cumsum(counts) .- counts`, a `Vector{Int}`, so comparing it to `0` is `false` for any input and the guard never fires. `counts_sum`, computed on the line above, is what was meant. Correcting only the condition would not be enough either: the branch would log and then fall out of the function returning `nothing`, which the caller assigns straight into `cache.nodes[1].is`.

The path is reachable. `cond` is `round(UInt8, 255 * rowsample)`, so any rowsample below 0.00196 keeps only mask bytes equal to zero, roughly 1 in 256 rows, and `check_args` permits any rowsample above `eps`. On small data the draw comes up empty routinely, and the fit currently completes with no splits and predicts the bias for every row, with nothing said.

In `init_core` an unsupported target eltype is reported the same way, then execution continues to `K = length(target_levels)` with `target_levels` still `nothing`. The helpful message is only a log line and the exception the user actually sees is unrelated:

```
[ Error: Invalid target eltype: Float64
MethodError: no method matching length(::Nothing)
```

This also masks the clearer single-level-target error added in #349 two lines further down.

The two in `fit-utils.jl` are unreachable through the table API, since features are filtered to `Union{Real,CategoricalValue}` on auto-detection and asserted against the same union when passed explicitly. They are converted anyway so the idiom is not copied again.

Adds tests in `test/core.jl` for the empty subsample and the invalid target eltype, and that a normal rowsample is unaffected. On the current code the first throws nothing at all and the second throws `MethodError` rather than `ErrorException`.

### DIFF
--- a/src/fit-utils.jl
+++ b/src/fit-utils.jl
@@ -48,7 +48,7 @@ function get_edges(df; feature_names, nbins, rng=Random.MersenneTwister(), kwarg
             featbins[j] = length(edges[j]) + 1
             feattypes[j] = true
         else
-            @error "Invalid feature eltype: $(feature_names[j]) is $(eltype(col))"
+            error("Invalid feature eltype: $(feature_names[j]) is $(eltype(col))")
         end
         if length(edges[j]) == 1
             edges[j] = [minimum(col)]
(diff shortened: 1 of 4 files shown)


jeremiedb:
Thanks for the fix and added tests!

jeremiedb (approved):

PR #370, opened 2026-08-26
Dev

jeremiedb:


AdityaPandeyCN:
@jeremiedb I see training speed for `xgb` became : ~2x on GPU for the 10M x 100 configs . Did you test it with some different configuration(algorithms...) or it was just pulling a newer version ?

jeremiedb:
I updated with a different hardware, but consistent with the latest EvoTrees benchmark runs and on latest XGBoost version (v3.4). It points that there must be further speed up opportunities. At least for mse regression, I think XGBoost only accumulate 2 rather than 3 values, since for mse, the hessian is a constant, hence the full computation only require gradient and weight. 
For in general, all of grad, hessian and weight are needed, so I'm having an issue to have more compute here in order to have a clean weight vs hessian tracking. 

One aspect might be the Float64 accumulation for histograms. Not sure if XGBoost is full Float32, but I know these could be subject to numerical instanbility for accumulation. If using Float32, it may benefit both RAM occupancy and speed.

AdityaPandeyCN:
> One aspect might be the Float64 accumulation for histograms. Not sure if XGBoost is full Float32, but I know these could be subject to numerical instanbility for accumulation. If using Float32, it may benefit both RAM occupancy and speed.

I see it has something different, Gradients are converted to fixed-point Int64 each round with a scale derived from the signed gradient sums, accumulated as integers, and converted back to double for split evaluation. So the stability question is handled by exactness rather than float width. CPU hist stays double.

Type: https://github.com/dmlc/xgboost/blob/v3.4.1/include/xgboost/base.h#L253
Scale from sum(pos), sum(neg): https://github.com/dmlc/xgboost/blob/v3.4.1/src/tree/gpu_hist/quantiser.cu#L55-L83
Encode, including the hess==0 -> 1 clamp: https://github.com/dmlc/xgboost/blob/v3.4.1/src/tree/gpu_hist/quantiser.cuh#L59-L69

This would be an interesting avenue to investigate and implement in Evotrees.jl. I am currently looking if I can do a minimal implementation of this.

PR #375, opened 2026-08-29
Ranking: group support, per-query NDCG and a LambdaRank objective

AJ0070:
Adds group (query) support for ranking on CPU and GPU: a group index through both `fit` entry points, group-aware `rowsample`, per-query NDCG, a LambdaRank objective, and a grouped correlation metric.

Groups are a per-row id column, `group_name` from a table or `group_train` / `group_eval` from matrices. `metric=:ndcg` scores within each group and averages over them, so early stopping tracks ranking quality instead of squared error. `loss=:lambdarank` weights pairs within a query by the NDCG change a swap would cause. The lambdas are per document, so K = 1 and the `2K+1` histogram, split gain and leaf solver are all untouched; the objective is one `update_grads!` method.

Groups are ids rather than XGBoost's `group_ptr_` boundaries, so they can arrive unsorted, non-contiguous or non-numeric and the rows need no pre-sorting. `rowsample` takes whole groups, since splitting one changes the comparison set the objective is computed over. A group with no relevant document scores 1.0, matching the `idcg == 0 ? 1.0 : ...` already in the LTRC tutorial.

`eval_group_name` sets the group column for `deval` alone and defaults to `group_name`, so a model can train with ordinary per-row sampling while a group-aware metric is tracked over queries.

A group's weight is the mean of its rows' weights, so unit weights leave every query equally weighted, matching the per-group weights ranking libraries take. Only that group-level weight reaches `:ndcg`, whose definition is a ranking of the group's documents. Where per-document weights should count, `metric=:corr` scores a weighted Pearson correlation within each group and uses weights at both levels. Groups carrying no signal, under two rows or a constant target, are left out of its average.

NDCG and the lambdas are computed host-side on GPU rather than as a segmented sort over variable-length groups. That keeps the two backends identical rather than merely close, and a device kernel is straightforward to add later if a profile ever asks for one.

On the gradient: holding the NDCG deltas fixed, it matches a finite difference of the pairwise cost to 1e-8 at two truncation levels, hessians are non-negative, and the gradients within a query sum to zero. GPU and CPU come out bit-identical.

Ranking only works through `EvoTrees.fit`. The MLJ interface has no way to pass groups, and the docstrings now say so.

Tests in `test/ranking.jl` cover the index, NDCG against a hand computation, the correlation metric against an independently written weighted Pearson, sampling keeping groups whole, the lambdarank gradient, and both entry points. `benchmarks/Yahoo-LTRC.jl` gains a lambdarank section alongside the existing ones, and `benchmarks/pythoncall/ranking.jl` compares both libraries on identical data and groups across a group-size sweep.

### DIFF
--- a/ext/EvoTreesKernelAbstractionsExt/metrics.jl
+++ b/ext/EvoTreesKernelAbstractionsExt/metrics.jl
@@ -131,11 +131,9 @@ function EvoTrees.mlogloss(p::CuMatrix{T}, y::CuVector, w::CuVector{T}, eval::Cu
     return sum(eval) / sum(w)
 end
 
-# NDCG needs the documents of each group sorted by predicted score. On device that means a
-# segmented sort over variable-length groups, which is substantially more machinery than
-# every other GPU metric here. NDCG is an eval metric computed once per boosting round on the
-# eval set rather than a hot path, so the predictions are brought to the host and scored
-# there. The result is identical to the CPU path by construction, since it is the same code.
+# NDCG needs each group sorted by predicted score, which on device means a segmented sort
+# over variable-length groups. It is an eval metric run once per round rather than a hot
+# path, so predictions are brought to the host and scored with the CPU code.
 function EvoTrees.ndcg(
     p::CuMatrix{T},
     y::CuVector,
--- a/ext/EvoTreesKernelAbstractionsExt/metrics.jl
+++ b/ext/EvoTreesKernelAbstractionsExt/metrics.jl
@@ -130,3 +130,27 @@ function EvoTrees.mlogloss(p::CuMatrix{T}, y::CuVector, w::CuVector{T}, eval::Cu
     KernelAbstractions.synchronize(backend)
     return sum(eval) / sum(w)
 end
+
+# NDCG needs the documents of each group sorted by predicted score. On device that means a
+# segmented sort over variable-length groups, which is substantially more machinery than
+# every other GPU metric here. NDCG is an eval metric computed once per boosting round on the
+# eval set rather than a hot path, so the predictions are brought to the host and scored
+# there. The result is identical to the CPU path by construction, since it is the same code.
+function EvoTrees.ndcg(
+    p::CuMatrix{T},
+    y::CuVector,
+    w::CuVector{T},
+    eval::CuVector{T};
+    group=nothing,
+    ndcg_k::Int=typemax(Int),
+    kwargs...
+) where {T<:AbstractFloat}
+    isnothing(group) && error(
+        "`metric = :ndcg` requires group information. Pass `group_name` when fitting from a " *
+        "table, or `group_eval` alongside `x_eval` when fitting from a matrix."
+    )
+    p_cpu = Array(p)
+    y_cpu = Array(y)
+    w_cpu = Array(w)
+    return EvoTrees.ndcg(p_cpu, y_cpu, w_cpu, similar(w_cpu, 0); group, ndcg_k)
+end
--- a/ext/EvoTreesKernelAbstractionsExt/metrics.jl
+++ b/ext/EvoTreesKernelAbstractionsExt/metrics.jl
@@ -157,7 +157,7 @@ end
 # arrays are brought to the host rather than scalar-indexed on device.
 function EvoTrees.corr(
     p::CuMatrix{T},
-    y::CuVector,
+    y::Union{CuVector,CuMatrix},
     w::CuVector{T},
     eval::CuVector{T};
     group=nothing,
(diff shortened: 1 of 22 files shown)


ext/EvoTreesKernelAbstractionsExt/metrics.jl
```diff
@@ -130,3 +130,25 @@ function EvoTrees.mlogloss(p::CuMatrix{T}, y::CuVector, w::CuVector{T}, eval::Cu
     KernelAbstractions.synchronize(backend)
     return sum(eval) / sum(w)
 end
+
```
jeremiedb, line 133: `corr` doesn't appear supported for GPU. 
And if attempting to use `corr` metric on a gpu run, it fails with:
```
ERROR: TaskFailedException
    nested task error: Scalar indexing is disallowed.
```

src/metrics.jl
```diff
@@ -142,6 +142,118 @@ function multiquantile(
...
+        "`metric = :ndcg` requires group information. Pass `group_name` when fitting from a " *
+        "table, or `group_eval` alongside `x_eval` when fitting from a matrix."
+    )
+    ng = ngroups(group)
+    scores = zeros(Float64, ng)
+    weights = zeros(Float64, ng)
+    @threads for g in 1:ng
+        rows = group_rows(group, g)
+        pred = [p[1, r] for r in rows]
+        rel = [y[r] for r in rows]
```
jeremiedb: are these allocating construction necessary? Assuming that groups are contiguous rows, couldn't views over range be preferable?

src/metrics.jl
```diff
@@ -142,6 +142,118 @@ function multiquantile(
...
+        "when fitting from a table, or `group_eval` alongside `x_eval` when fitting from a matrix."
+    )
+    ng = ngroups(group)
+    scores = zeros(Float64, ng)
+    weights = zeros(Float64, ng)
+    @threads for g in 1:ng
+        rows = group_rows(group, g)
+        pred = [Float64(p[1, r]) for r in rows]
+        obs = [Float64(y[r]) for r in rows]
+        wt = [Float64(w[r]) for r in rows]
```
jeremiedb: Same comment as above for ndcg: are tese constructors needed?
Why convert to Float64? Should be let to type `T` (Float32).

src/metrics.jl
```diff
@@ -142,6 +142,118 @@ function multiquantile(
...
+`eval_group_name` or `group_eval`. Unlike `:ndcg` this uses weights at both levels: a row's
+weight enters its own group's correlation, and a group weighs by the mean of its rows' weights.
+
+Groups of fewer than two rows, and groups whose target is constant, carry no signal and are
+left out of the average. A group whose prediction is constant while its target is not scores
+zero.
+"""
+function corr(
+    p::AbstractMatrix{T},
+    y::AbstractVector,
```
jeremiedb: Could multitarget be supported here? Ie: `y::AbstractVecOrMat{T}`?
In any case, `y` can be enforced to be `T`

jeremiedb (commented):
Just few comments from initial review.

ext/EvoTreesKernelAbstractionsExt/metrics.jl
AJ0070: Fixed. Added the host-copy method alongside `ndcg`'s, same reasoning. On a T4 `:corr` now returns the same value on cpu and gpu.

src/metrics.jl
AJ0070: Groups are row ids rather than boundaries, so a group's rows need not be contiguous: sorted input gives contiguous rows, interleaved gives stride 1000. Views over a range don't hold in general.

Most of it wasn't the gather though. The `@threads` closure was boxing its captures, so every per-group call was a dynamic dispatch. Function barrier, reused scratch and in-place `QuickSort`: 9.90 -> 0.24 MiB at group size 20, 9.80 -> 0.06 at 100. Time is about the same.

src/metrics.jl
AJ0070: Same fix here: 17.32 -> 0.23 MiB at group size 20 and 14.21 -> 0.06 at 100, and roughly 5x faster since the dispatch went away.

Signature takes `T` now, but the accumulator stays Float64: the centring cancels in Float32 once predictions sit far from zero, 3e-2 error at offset 1e7. Related, `mean(w[r] for r in rows)` was returning Float32, so the group weight carried 1e-7. Metric values shift ~2e-10 as a result, identical with unit weights.

src/metrics.jl
AJ0070: Done, `y::AbstractVecOrMat{T}`. Each target correlates on its own and the group takes their mean, like `_eval_metric` does with `acc / K`.

One catch: K is the target count rather than `size(p, 1)`. For `gaussian_mle` row 2 is the scale, and averaging it in halved the score.

jeremiedb:
@AJ0070 Looks good and ready for merge! Any pending item you see on your end?

AJ0070:
One thing, since it is the same point you raised on the metrics: the per-group allocation was also in the lambdarank gradient, which runs over the training set every round rather than once on eval. Same fix there, 16.24 -> 1.24 MiB at group size 20 and 14.59 -> 0.29 at 100. Gradients are bit-identical to before, so nothing about the fit changes.

Nothing else pending on my end.

jeremiedb (approved):

PR #380, opened 2026-08-31
Reuse train bins for eval and binarize GPU matrices on device.

AdityaPandeyCN:
I aim to remove the CPU `binarize` cost from GPU fit and predict.

- `fit`: when `x_eval === x_train` (or `deval === dtrain`), pass `cache.x_bin` to the callback instead of binning again. Also drops the duplicate `x_bin` on device. Applies to both backends.
- `binarize(device, data; feature_names, edges)`: new device-dispatched entry. CPU forwards to the existing `binarize`. GPU bins a dense `Matrix` on the device one column at a time (same `searchsortedfirst` broadcast as the CPU path, extra device memory = one column); Tables and other inputs still go through CPU `binarize` then copy. Used in GPU `init_core`, `_predict`, `_predict_leaf_indices`, and the eval callback.


AdityaPandeyCN:
@jeremiedb Please test the benchmark performance from your end once you have time. I will add more changes but wanted to test this idea for now.

jeremiedb (approved):
Looks clean, thanks!

PR #386, opened 2026-09-06
Score each target against its own location in the grouped corr metric

AJ0070:
The grouped `corr` metric reads prediction row `k` for target `k`, but a two parameter likelihood interleaves location and scale, so target `k` predicts into row `2k - 1`. With two targets the second one is therefore correlated against the first target's unconstrained scale rather than against its own location.

On 20k rows in groups of 64, `EvoTreeMLE(loss=:gaussian_mle, metric=:corr)` with two targets:

```
metric as logged                            0.4595
same computation over the location rows     0.9885
target 2 against row 2, target 1 log scale -0.0660
target 2 against row 3, target 2 location   0.9925
```

Fixing my own oversight from #375. A single target is unaffected, which is why it went unnoticed: `K` is 1 there and row 1 is the location either way, so the existing coverage passed.

The layout is read off the prediction matrix rather than passed in with the loss, so the metric keeps the loss agnostic interface you asked for on #327.

Worth flagging that `benchmarks/regressor-grps.jl` sets `n_targets = 2` with `gaussian_mle` and `corr`, so the committed grouped results were tracking this.

Tests cover the interleaved layout directly and a two target fit. Both fail on current main.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -167,7 +167,7 @@ function _ndcg_chunk!(scores, weights, p, y, w, group, chunk, ndcg_k::Int)
     return nothing
 end
 
-function _corr_chunk!(scores, weights, p, y, w, group, chunk, K::Int)
+function _corr_chunk!(scores, weights, p, y, w, group, chunk, K::Int, stride::Int)
     pred = Float64[]
     obs = Float64[]
     wt = Float64[]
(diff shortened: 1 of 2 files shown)


jeremiedb:
Good catch, thanks for paying attention to these per-group implications.

PR #394, opened 2026-09-10
Jdb/v20

jeremiedb:


AJ0070:
`load` is unchanged on this branch:

```julia
function load(path)
    m = BSON.load(path, @__MODULE__)
    return m[:model]
end
```

`EvoTree` gained `bias` in #384 and `trees` no longer holds the intercept, so BSON rebuilds the old
layout by field position and the previous `trees` vector lands in `bias`. Fitting and saving under
the registered v0.19.0, then loading on this branch, fails with a `MethodError` converting
`Memory{Any}` to `Memory{Float32}`. It applies to every version released so far, not just v0.19.

#387 recognises the old layout in `load` and converts it, and stamps a version into `info` on save so
the next structure change has something to branch on. It is green and rebased.

Worth landing before the tag rather than after, since once v0.20 is out it stops being a
pre-release fix and becomes a regression for anyone with a saved model.

PR #402, opened 2026-09-19
Add optional per-round callbacks to fit

sylvainpronovost:
## Summary

Add an optional `callbacks=()` keyword to both `fit` interfaces. This provides a small, framework-neutral hook for progress reporting, instrumentation, and externally controlled stopping without changing the default training behavior or adding dependencies.

Each callable receives `callback(model, logger, iteration)` once per completed boosting round, after the validation update and before the native early-stopping check. The iteration-zero baseline is not emitted. Without evaluation data, `logger` is `nothing`.

Callback return values are ignored and exceptions propagate. Callbacks are documented to treat the model and logger as read-only. The existing internal validation callback remains responsible for computing metrics; this hook exposes the completed-round boundary to callers.

## Scope

Only the two fit entrypoints, their docstrings, and focused regression tests change. No dependency, serialization, MLJ update, accelerator, or model-training algorithm changes.

## Validation

The focused callback testset passes **36/36 assertions** on Julia 1.12.7 / CPU. It covers both table and matrix interfaces, callback order, bagging round counts, finite validation metrics, unchanged predictions with empty/disabled callbacks, exception propagation, missing evaluation data, and callback precedence when native early stopping also fires.

```julia
using Test, Random, EvoTrees
using EvoTrees: fit, predict
include("test/fit_callbacks.jl")
```

This is focused CPU evidence, not a claim of a complete accelerator test run.


sylvainpronovost:
Additional local CPU regression check: test/core.jl, test/tables.jl and test/metrics.jl passed 285 assertions on Julia 1.12.7, alongside the 36 focused callback assertions. The contribution remains limited to the three files in the original diff.

